PR #2, opened 2022-02-10
feat: add new polyfill for event source 

RitaDias:
EventSource/SSE which we are using for listeners in the Studio will not accept headers, and thus can't be used with token auth.

We need to use a polyfill for this. The client was already using one, however, the polyfill wasn't working well for Safari so we had to change it.

This PR is the end result of a spike that was done, and tried to use https://github.com/EventSource/eventsource and it works very well! 

We changed polyfill from the first one we had considered because it caused a lot of conflicts with the listener tests that were already set up.

We've made the decision to only ever use the polyfil (vs checking if the browser has a default EventSource) so that it will be more consistent in the future and easier to debug.

### DIFF
--- a/src/data/listen.js
+++ b/src/data/listen.js
@@ -1,9 +1,7 @@
 const assign = require('object-assign')
 const {Observable} = require('../util/observable')
 const polyfilledEventSource = require('@sanity/eventsource')
-const generateHelpUrl = require('@sanity/generate-help-url')
 const pick = require('../util/pick')
-const once = require('../util/once')
 const defaults = require('../util/defaults')
 const encodeQueryString = require('./encodeQueryString')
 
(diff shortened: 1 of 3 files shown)


src/data/listen.js
```diff
@@ -20,9 +20,7 @@ const tokenWarning = [
 const printTokenWarning = once(() => console.warn(tokenWarning.join(' ')))
```
skogsmaskin: This one should now be removed as we actually supports tokens now.

skogsmaskin (changes requested):

src/data/listen.js
RitaDias: so In that case we should remove https://github.com/sanity-io/client/blob/4719587794b5a73a3356dd99a818344bb26c179d/src/data/listen.js#L15-L18

and 
https://github.com/sanity-io/client/blob/4719587794b5a73a3356dd99a818344bb26c179d/src/data/listen.js#L52-L54 

as well? Since that's where that line is used? :)

skogsmaskin: Yes! 👍

rexxars (approved):
👍  Looks good! Tried it out locally, seems to work as expected. Should have proper multi-browser testing inside the studio context - maybe we can do a tagged release for now?

PR #9, opened 2022-03-22
feat: add `dryRun`, `autoGenerateArrayKeys` mutation options

### LINKED ISSUE #6: Use dryRun option
Can the option to use dryRun for mutations be added?

[dryRun documentation](https://www.sanity.io/docs/http-mutations#dryRun-cd0dae296271)

rexxars:
This PR adds two missing mutation options to the client mutation options:
 
- `dryRun` (`true|false`) - default `false`. If true, the mutation will be a dry run - the response will be identical to the one returned had this property been omitted or false (including error responses) but no documents will be affected.
- `autoGenerateArrayKeys` (`true|false`) - default `false`. If true, the mutation API will automatically add `_key` attributes to objects in arrays that is missing them. This makes array operations more robust by having a unique key within the array available for selections, which helps prevent race conditions in real-time, collaborative editing.

Also slightly tightens up the readme to be a bit clearer on the availability of the mutation options.

Closes #6

### DIFF
--- a/src/data/dataMethods.js
+++ b/src/data/dataMethods.js
@@ -14,12 +14,12 @@ const excludeFalsey = (param, defValue) => {
 
 const getMutationQuery = (options = {}) => {
   return {
+    dryRun: options.dryRun,
     returnIds: true,
     returnDocuments: excludeFalsey(options.returnDocuments, true),
     visibility: options.visibility || 'sync',
-    ...(options.skipCrossDatasetReferenceValidation && {
-      skipCrossDatasetReferenceValidation: options.skipCrossDatasetReferenceValidation,
-    }),
+    autoGenerateArrayKeys: options.autoGenerateArrayKeys,
+    skipCrossDatasetReferenceValidation: options.skipCrossDatasetReferenceValidation,
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/data/dataMethods.js
```diff
@@ -14,12 +14,12 @@ const excludeFalsey = (param, defValue) => {
 
 const getMutationQuery = (options = {}) => {
   return {
+    dryRun: options.dryRun,
     returnIds: true,
     returnDocuments: excludeFalsey(options.returnDocuments, true),
     visibility: options.visibility || 'sync',
-    ...(options.skipCrossDatasetReferenceValidation && {
-      skipCrossDatasetReferenceValidation: options.skipCrossDatasetReferenceValidation,
```
skogsmaskin, line 21: I see this was removed here previously. Would it make sense to remove the undefined ones still (if they end up being part of the query - in order to compress request URLs)?

skogsmaskin (approved):
Looks good to me 👍

src/data/dataMethods.js
rexxars: It automatically drops `undefined` values, while boolean parameters are set to `true`/`false` - so I think the change makes it easier to read.

PR #11, opened 2022-04-24
feat: adding ESM support

stipsan:
BREAKING CHANGE:
There are no longer any default exports, only named ones.

What we reap from this is the ability to run on systems like skypack and deno out of the box.
I'm opening this as a draft PR for now but will circle back and follow up later.


stipsan:
Superseded by #13

PR #13, opened 2022-06-08
feat: add ESM export

stipsan:
Adds `module` and `exports` to `package.json`.
The files referenced are generated using `esbuild`.

# What this solves
A lot of new, modern, environments require packages to start shipping ESM.
For example when attempting to use `@sanity/client` in a [Remix](https://remix.run/) site will yield an error like the one below:
image
This is from `same-origin`, which is included by `get-it`, which is used by `@sanity/client`.
To solve this there's a new build command `npm run esbuild:browser` that transforms CJS to ESM and includes them in the bundle. This solves the problem, for now, buying us time to migrate these packages to ESM:
- https://www.npmjs.com/package/@sanity/eventsource
- https://www.npmjs.com/package/@sanity/generate-help-url
- https://www.npmjs.com/package/get-it
- https://www.npmjs.com/package/make-error
- https://www.npmjs.com/package/rxjs

Once they are shipping ESM we can stop prebundling them:
```diff
    "build": "npm run compile && npm run browserify && npm run minify && npm run esbuild:browser && npm run esbuild:node",
-   "esbuild": "esbuild src/sanityClient.js --bundle --sourcemap",
+   "esbuild": "esbuild src/sanityClient.js --bundle --sourcemap --external:rxjs --external:@sanity/eventsource --external:get-it --external:make-error --external:@sanity/generate-help-url",
    "esbuild:browser": "npm run esbuild -- --format=esm --outfile=dist/sanityClient.browser.mjs --platform=browser",
-   "esbuild:node": "npm run esbuild -- --format=cjs --outfile=dist/sanityClient.node.cjs --platform=node --external:object-assign --external:rxjs --external:@sanity/eventsource --external:get-it --external:make-error --external:@sanity/generate-help-url",
+   "esbuild:node": "npm run esbuild -- --format=cjs --outfile=dist/sanityClient.node.cjs --platform=node",
    "lint": "eslint .",
```

## How to test
A test version is published under `npm i @sanity/client@esm`.

### Using unpkg.com
Run this in your console (not on github as it blocks external scripts, but for example on https://sanity.io or a localhost):
```js
const {default: createClient} = await import('https://unpkg.com/@sanity/client@esm?module=true')
createClient() // Should throw as no projectId is given
```

### Using skypack.dev
```js
const {default: createClient} = await import('https://cdn.skypack.dev/@sanity/client@esm')
createClient() // Should throw as no projectId is given
```

## `./dist/sanityClient.browser.mjs`
This file is generated to run in ESM strict envs, like deno, native browser ESM using CDNs like unpkg or skypack. To do this it bundles everything (no `import` or `require` calls left) and it applies our `pkg.browser` field making sure to load `browserMiddleware.js` instead of `nodeMiddleware.js`.

## `./dist/sanityClient.node.cjs`
Generated to run in allows calling `require`, using `.cjs` to make it clear to modern 'node' that it's a CommonJS file.

Once we have a bundler that implements [module.createRequire](https://nodejs.org/api/module.html#modulecreaterequirefilename) we could consider adding a `./dist/sanityClient.node.mjs` version.

## `pkg.module`
This only affects bundlers that are modern enough to understand what it does, and does not trigger modern node to run in ESM module (unlike `pkg.type: "module"`). It points to `dist/sanityClient.browser.mjs` instead of `dist/sanityClient.node.js` as it has the highest chance of bundling success since `pkg.browser` is already applied to it.

## adding `package-check` to `posttest`
Runs [@skypack/package-check](https://github.com/skypackjs/package-check) to validate we're following ESM best practices.


.eslintrc
```diff
@@ -8,6 +8,6 @@
     "browser": true
   },
   "parserOptions": {
-    "ecmaVersion": 2018
+    "ecmaVersion": 2020
```
stipsan, line 8: Fixes a parse error due to the dynamic `import()` in `client.test.js`

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x, 18.x]
```
stipsan, line 11: node v12 is dead, longe live node v12 👑

PR #19, opened 2022-08-15
chore: add runtime tests

stipsan:
Adds tests for new runtimes: [Bun](https://bun.sh/), [Deno](https://deno.land/) and the [Edge runtime](https://edge-runtime.vercel.sh/). I wanted to add tests for node16 ESM modes as well as Cloudflare Workers and react-native maybe. But I'm out of time and it's best to get in what we have and follow up with new PRs.

The Deno test is expected [to keep failing](https://github.com/sanity-io/client/runs/7849905824?check_suite_focus=true#step:5:201) until [`get-it` gets a similar](https://github.com/sanity-io/get-it/blob/7873e3fb4f38c21fe91f727f65acca250bc6c5f0/package.json#L9-L23) fix as #24

Let me know if you prefer me to comment out the Deno workflow until the test is passing.


rexxars (approved):
Gracias! 

I'm guessing we need to remember to update the bun lockfile whenever dependencies are updated?

rexxars:
Also, yes - I think we should disable the deno test for now until we can get it working.

PR #21, opened 2022-08-16
chore: add type checking script

runeh:
Adds typescript to verify the typings file as part of running the tests.


rexxars (approved):
Thanks!

PR #23, opened 2022-08-18
Improve wording for API version in readme

ledenis:
I find this sentence confusing:

> In general, unless you know what API version you want to use, you'll want to set it to today's UTC date.

I understand it as setting the date dynamically to today's UTC date, which is not the intended meaning. This PR suggests an improvement on the wording

I also submitted the same suggestion to this article : https://www.sanity.io/help/js-client-api-version


rexxars (approved):
Thanks!

PR #24, opened 2022-08-19
feat: support modern bundlers browser mode

stipsan:
This is based on #13 but far less ambitious.
- There is no [aggressive](https://github.com/sanity-io/client/blob/4ab7b480b7a7cc1ce992e66e0598cc00c44ee43a/package.json#L28) bundling (`get-it`, `rxjs` etc stays external).
- No [massive tests rewrite](https://github.com/sanity-io/client/blob/4ab7b480b7a7cc1ce992e66e0598cc00c44ee43a/package.json#L37).
- No [`node` specific `pkg.exports`](https://github.com/sanity-io/client/blob/4ab7b480b7a7cc1ce992e66e0598cc00c44ee43a/package.json#L13).

What is the point of this PR then?
It's to add a `pkg.exports.browsers` so that browser targeted bundling works in modern bundlers that [doesn't support `pkg.browser`](https://github.com/sanity-io/client/blob/4ab7b480b7a7cc1ce992e66e0598cc00c44ee43a/package.json#L40-L43).

What about ESM?
This PR improves that story slightly, not because the generated `sanityClient.browser.mjs` is `ESM` formatted, but only because `esbuild` that generates it understands `pkg.browsers` and makes sure to use [`src/http/browserMiddleware.js` instead of `src/http/nodeMiddleware.js`](https://github.com/sanity-io/client/blob/4ab7b480b7a7cc1ce992e66e0598cc00c44ee43a/package.json#L40-L43).
This means that bundlers that are currently failing to build `@sanity/client` due to it attempting to[ bundle `get-it/lib-node/middleware/*` and its dependencies](https://github.com/sanity-io/client/blob/428bc53ab4d9879baf466030e0fc9e048e9eb9ef/src/http/nodeMiddleware.js#L1-L3), such as skypack:
```
await import('https://cdn.skypack.dev/@sanity/client')
// GET https://cdn.skypack.dev/-/speedometer@v1.0.0-7ZqxVFTYH3353ZTAOfaN/dist=es2019,mode=raw/speedometer net::ERR_ABORTED 404
```
will now start working again with this PR.

I'll revisit #13 once all the upstream deps are ESM. That includes the `node` part of it as I need more time testing node 16 and its different modes with the new `pkg.exports.node` to fully understand how it works and how we should wield it.


rexxars:
Thanks a bunch!

PR #26, opened 2022-08-25
chore: let robots maintain `bun.lockb`

stipsan:
> I'm guessing we need to remember to update the bun lockfile whenever dependencies are updated?
> https://github.com/sanity-io/client/pull/19#pullrequestreview-1084438726

Let's not need to remember doing that 😌


rexxars (approved):
This is great. Thank you!

PR #29, opened 2022-08-26
fix: improve ESM output to support Deno, Bun and Edge runtimes

stipsan:
Supersedes #13.
Based on [new research](https://client-runtimes-dashboard.sanity.build/) and findings on how conditional package exports actually works in Bun, Deno, Cloudflare Workers and Edge runtimes.

~~Blocked by https://github.com/sanity-io/get-it/pull/25~~

Testable by
```bash
npm i @sanity/client@esm
```
As well as
```js
const {default: createClient} = await import('https://esm.sh/@sanity/client@esm')
```
```js
const {default: createClient} = await import('https://cdn.skypack.dev/@sanity/client@esm')
```


rexxars (approved):
Nice work! Have not tested this in all environments, but trust that you have and the tests we have in place are sufficient :)

Still think we should release this as a major version, just in case.

PR #32, opened 2022-09-05
fix: Add params to MutationSelection

### LINKED ISSUE #31: MutationSelection types doesn't match documentation
According to the documentation on deletion [here](https://www.sanity.io/docs/js-client#delete-documents) one should be able to do this:
```js
client
  .delete({query: '*[_type == $type][0..1]', params: {type: 'bike'}})
  .then(() => {
    console.log('The documents matching *[_type == "bike"][0..1] was deleted')
  })
  .catch((err) => {
    console.error('Delete failed: ', err.message)
  })
```

But the [`MutationSelection`](https://github.com/sanity-io/client/blob/3c51f654d9e21d9d5fd2e85931a62f796c2652ca/sanityClient.d.ts#L38) type, which is used by [delete](https://github.com/sanity-io/client/blob/3c51f654d9e21d9d5fd2e85931a62f796c2652ca/sanityClient.d.ts#L1382-L1392), is ``type MutationSelection = {query: string} | {id: string}`` , and therefore errors out.

Depending on where that type is used it should be as simple as:
```ts
type MutationSelection = {query: string} | {query: string, params: QueryParams } | {id: string}
// or
type MutationSelection = {query: string, params?: QueryParams} | {id: string}
```

albe2669:
Fixes: #31 

This PR simply adds `params?: QueryParams` to the `MutationSelection` type so that it matches the documentation.

The change should be fine as it both matches the documentation, and every method that uses this type uses the utility function `getSelection` which supports every case that the `MutationSelection` type now supports.

### DIFF
--- a/sanityClient.d.ts
+++ b/sanityClient.d.ts
@@ -35,7 +35,7 @@ export interface ProgressEvent {
 
 type AttributeSet = {[key: string]: any}
 type QueryParams = {[key: string]: any}
-type MutationSelection = {query: string} | {id: string}
+type MutationSelection = {query: string, params?: QueryParams} | {id: string}
 type SanityReference = {_ref: string}
 
 interface RawRequestOptions {


rexxars (approved):
Mange millioner tak 👏

PR #33, opened 2022-09-12
fix: use `NodeJS.ReadableStream` type for upload body type to prevent conflict with DOM `ReadableStream`

juice49:
Attempting to pass a node `ReadableStream` to the `upload` function currently results in a type error, because the client library inadvertently expects a DOM `ReadableStream`. This change to the types explicitly references node's `ReadableStream`, resolving the type error.


stipsan (approved):
Thnx! :relaxed:

PR #35, opened 2022-09-15
chore: update rxjs

SimenB:
Main reason is that I want to update the underlying `tslib` which causes deprecation warnings.

I haven't tried to run this or anything, so this PR is more of a "please upgrade" than a PR that's actually good to go 😅 

https://rxjs.dev/deprecations/breaking-changes


bjoerge:
Superseded by #80 - thanks for the nudge, @SimenB :)

PR #51, opened 2022-10-21
fixed issue #9 url problem fixed

Hallcoder:


Hallcoder:
The url.parse method that was not working now works after installing the right dependency

stipsan:
Could you give us a repro case of when it doesn't work? :relaxed:

stipsan:
Closing this as the upcoming `v5` release solves this by using the [`Web URL API`](https://developer.mozilla.org/en-US/docs/Web/API/URL) for parsing, which doesn't need an npm dependency :relaxed:

Hallcoder:
That's great as it's resolved!

PR #78, opened 2023-01-02
build: replace browserify with rollup for UMD bundling

bjoerge:
This switches from browserify to rollup for generating UMD bundle. This should give us several improvements, most notably reduced bundle size due to tree shaking. We also need this to be able to upgrade rxjs to v7 since they have removed the ability to require internal individual files with commonjs.

Some quick stats:
Before (browserify):
127K ./umd/sanityClient.js
108K ./umd/sanityClient.min.js

After (rollup):
174K ./umd/sanityClient.js
70K  ./umd/sanityClient.min.js

Note: I've only done a quick round of testing the bundle loaded through a script tag and it seems to work well. Let me know if there's a way to run a more extensive test suite against it.


stipsan (approved):
Great stuff!!

rexxars (approved):
👍

PR #80, opened 2023-01-04
fix(deps): upgrade rxjs to v7

bjoerge:
This upgrades rxjs from v6 to v7. This means that the observable instance returned from various client methods will now be an rxjs@v7 `Observable`. This *should* be a non-breaking since the `Observable` interface and behavior is from what I can tell largely kept intact  between v6 => v7 (except for undocumented methods and a couple of minor type signature changes). In other words: a v6 Observable should be interoperable with a v7 Observable.

I set the range of the `rxjs` dependency to `^7.0.0` to aid deduping, and verified that the test suite passes against every minor of rxjs from `v7.0.0` to `v7.8.0`.

Note: would prefer if the PR was rebase-merged since each individual commit makes sense on their own :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@sanity/eventsource": "^4.0.0",
         "get-it": "^7.0.2",
         "make-error": "^1.3.6",
-        "rxjs": "^6.6.7"
+        "rxjs": "^7.0.0"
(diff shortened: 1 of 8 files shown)


stipsan (approved):
Good to go 👍

PR #83, opened 2023-01-06
refactor: remove usage of deprecated rxjs methods

bjoerge:
just a quick follow-up from #80 as I realized that we had more usage of the now deprecated `<observable>.toPromise()`


stipsan (approved):
Good catch!

PR #86, opened 2023-01-09
feat: full Node.js ESM runtime support

stipsan:
Squash merge commit body:
```
BREAKING CHANGE: We have removed the default export and replaced it with a named one:

    ```diff
    -import SanityClient from '@sanity/client'
    +import {createClient} from '@sanity/client'
    ```
We're working on a codemod for this migration.
[The migration guide outlines every breaking change and how to migrate your code](https://github.com/sanity-io/client#from-v4)

- feat: full ESM support in modern runtimes and tooling (Bun, Deno, Edge, without breaking Node.js ESM and CJS compatibility)
- feat: codebase rewritten in TypeScript, typings are generated and no longer manually maintained
- feat: make `httpRequest` on `SanityClient` configurable, intended for libraries wishing to extend the client
- feat: shipping modern syntax, reducing bundlesize
```

# How to test
- `npm install @sanity/client@esm`
- browser (DevTools Console): `const {createClient} = await import('https://esm.sh/@sanity/client@esm')`
- umd: `<script src="https://unpkg.com/@sanity/client@esm"></script>`

The generated typings shouldn't diverge from the [original](https://github.com/sanity-io/client/blob/371515b9c517c1849fcd2fc214872fb7fa7e7061/sanityClient.d.ts). Maybe we can use `api-extractor` to compare the two `.d.ts` files and see exactly what's changed?

# Docs
 - [Formatted migration docs](https://github.com/sanity-io/client/tree/modernize#from-v4)
 - [Formatted setup docs for new runtimes](https://github.com/sanity-io/client/tree/modernize#creating-a-client-instance)

# Changes

- Codebase rewritten to TypeScript and modern JS syntax to make it easier to build new features.
- Generating types from source, no more manual type definitions.
- Replaced `@babel/cli` and `esbuild` tooling with `@sanity/pkg-utils`, ensuring the generated bundles, typescript definitions and `pkg.exports` are following our best practices.
- The new ESM setup brings us to [100% coverage for our runtimes target](https://client-runtimes-dashboard.sanity.build/), tested with `npm i @sanity/client@esm` end-to-end.
- Upgraded to `get-it` `v8` which ships with the same ESM support and modern syntax as this PR.
- Deprecated APIs are removed, with migration instructions in the readme. Some APIs are still marked as `@deprecated`, but they're actually "internals" that other core libraries we maintain rely on. Marking them `@deprecated` will in most IDE's create a strikethrough marking, to help surface to the third party that they're using an API that could have breaking changes in any release.
- Refactored internals to use [private class properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) and [closures](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures#emulating_private_methods_with_closures). What remains are public APIs and internal ones that other libraries, like `sanity`, rely on.
- Replaced the `pkg.browser` setup with the same setup `get-it` uses for conditionally loading dependencies that use Node APIs (`process.env`) or Web/Browser/Worker APIs (`location.url`). They generate a separate bundle for Node `src/index.ts => dist/index.js` and Browsers `src/index.browser.ts => dist/browser.js`. The previous setup with `pkg.browser` is only supported by bundlers like `browserify` and `webpack`. The new setup will load the browser or node optimal version of the client for you automatically if the environment supports `pkg.exports`. If `pkg.exports` isn't supported then you can fallback to manually `require('@sanity/client/dist/index.browser.cjs')` or `import('@sanity/client/dist/index.browser.js')` and there's still no bundler required.
- The new ESM support in Node.js doesn't sacrifice CJS compatibility. Wether you `require('@sanity/client')` or `import '@sanity/client'` is up to you.
- The ESM implementation is safe from the [`Dual Package Hazard`](https://nodejs.org/api/packages.html#dual-package-hazard) unique to the Node.js runtime. This means that checks like `client instanceof SanityClient` works even if `client` is loaded by CJS while `SanityClient` is an ESM import.
- `sideEffects: false` added to `@sanity/client/package.json` and `get-it/package.json` to enable much better tree-shaking dead code elimination in bundlers like webpack. 
- Bundlesize reductions:
	- Before
		- [Bundlephobia](https://bundlephobia.com/package/@sanity/client@4) ![minified](https://img.shields.io/bundlephobia/min/@sanity/client@4?label=minified&style=flat-square) ![minified+gzip](https://img.shields.io/bundlephobia/minzip/@sanity/client@4?label=minified+gzip&style=flat-square)
	- After
		- [Bundlephobia](https://bundlephobia.com/package/@sanity/client@esm) ![minified](https://img.shields.io/bundlephobia/min/@sanity/client@esm?label=minified&style=flat-square) ![minified+gzip](https://img.shields.io/bundlephobia/minzip/@sanity/client@esm?label=minified+gzip&style=flat-square)

# Internal changes (context for the team)
- Replaced custom build tooling with `@sanity/pkg-utils`, and applied its best practice for dealing with ESM interop with Node.js, bundlers and runtimes.
- Replaced `tape` with `vitest`.
- Tests are refactored to TypeScript to aid quality assuring the typings we generate, increasing their accuracy.
- Replaced `istanbul` for codecoverage metrics to `c8`, as it's using the built in Node.js coverage tools and have better support for esm.
- The migration to TypeScript used [sanityClient.d.ts](https://github.com/sanity-io/client/blob/371515b9c517c1849fcd2fc214872fb7fa7e7061/sanityClient.d.ts) as point of reference, however it only contains public typings. That's why there's a liberal use of `as any`. Implicit `any` aren't allowed, they have to be explicit. This makes it easier to come back later and type them properly.
- Because of the amount of `any` warnings in ESLint the CI uses `--quiet` to hide them. Once the `any` types are taken care of the intention is to set the CI to use `--max-warnings 0` instead.
- The CI is setup to report failing tests as annotations. It's as if `vitest` leaves[ a code review on your PR ](https://user-images.githubusercontent.com/49056869/162130129-137511a1-9df4-4431-af88-5626d66f587a.png)and requests changes on the failing tests directly. Saving you the round trip of looking at CI logs and figuring out where the code is in your IDE. 
- The `vitest` runs the suite in different environments (`node`, `happy-dom`/browser-like, `edge-runtime`) with different `pkg.exports` conditions to try and cover as much ground as possible.
- The deno runtime test is refactored to use `import_map.json` without a `deno.lock` file since our renovatebot setup doesn't support `deno`. The new setup have a `deno.yml` action that, on commits pushed to main that touches `package.json`, will update the `import_map.json` to use the same semver range as the client `dependencies`. 
- There's a Node.js runtime suite (`npm run test:node-runtimes`) that uses the built-in test runner introduced in [v18](https://nodejs.org/docs/latest-v18.x/api/test.html). This suite tests that Node.js is able to import the client in both its CJS and ESM runtime without the help of a bundler. 
- The CI is setup to automatically run `npx prettier --write .` if commits are pushed to `main` that forgot to run it. Especially useful if renovatebot updates `prettier` to a new version with formatting changes, so that the next person who wants to change 1 line of code in a PR don't end up with `prettier` creating tons of unrelated changes on their PR.
- And finally, there's a `lock.yml` GitHub action that automatically locks issues and PRs once they're closed. We don't get notifications if someone posts a comment on a closed thread, and so by locking them we nudge people to create a new issue. Enabling us to follow up and triage promptly.

# Next steps
- Work through the `any` usage in the client, and `get-it`.
- Remove `extract.rules` from `package.config.ts` and use defaults, correct incorrect tags.
- Either reuse `@sanity/types`, or move them here so we have a single source of truth for things like `SanityDocument` and `SanityImageAsset`.
- Look into if `@sanity/pkg-utils` can handle the checks we're doing in `npm run test:node-runtimes` and `npm test exports.test.ts` as part of its `package.json` validation.
- Update our [docs](https://www.sanity.io/help/js-client-promise-polyfill) as we're still saying that the Client requires `Promise` to be polyfilled, now that we use native ESM as the baseline maybe we should explain that instead?
- Lazyload `@sanity/eventsource` instead of always bundling it , we do this in `@sanity/preview-kit` already for bundlesize reasons. And since it's only used with `client.listen`, which is async, we can do this without introducing a breaking change.


stipsan:
@kmelve you might want to give the readme changes a look 👀

README.md
```diff
@@ -1,26 +1,85 @@
 # @sanity/client
 
+[![npm stat](https://img.shields.io/npm/dm/@sanity/client.svg?style=flat-square)](https://npm-stat.com/charts.html?package=@sanity/client)
 [![npm version](https://img.shields.io/npm/v/@sanity/client.svg?style=flat-square)](https://www.npmjs.com/package/@sanity/client)
-
-Javascript client for Sanity. Works in node.js and modern browsers (older browsers need a [Promise polyfill](https://www.sanity.io/help/js-client-promise-polyfill)).
+[![gzip size][gzip-badge]][bundlephobia]
+[![size][size-badge]][bundlephobia]
+
+Javascript client for Sanity. Works in [Node.js](https://nodejs.org/en/), [Bun], [Deno], [Edge Runtime] and [modern browsers].
```
bjoerge: Nit while we're at it: should be Java**S**cript, not Java**s**cript

.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
bjoerge: What dependenc(y/ies) do we need this for?

README.md
```diff
@@ -1,26 +1,85 @@
...
+npm install @sanity/client
 ```
 
 ## API
 
+### Creating a client instance
+
+`const client = createClient(options)`
+
+Initializes a new Sanity Client. Required options are `projectId`, `dataset`, and `apiVersion`. Setting a value for `useCdn` is encouraged.
```
bjoerge: Could we link/explain why `useCdn` is encouraged?

src/http/errors.ts
```diff
@@ -1,23 +1,43 @@
...
+import {ErrorProps} from '../types'
+
+/** @public */
+export class ClientError extends BaseError {
+  response: ErrorProps['response']
+  statusCode: ErrorProps['statusCode'] = 400
+  responseBody: ErrorProps['responseBody']
+  details: ErrorProps['details']
+
+  constructor(res: any) {
```
bjoerge: why `any` here? If it's a temporary change until we get around to fix it properly maybe we could add a todo comment or alias type to signal that it's something we want to fix?

src/http/errors.ts
```diff
@@ -38,18 +58,13 @@ function extractErrorProps(res) {
   return props
 }
 
-function httpErrorMessage(res) {
+function httpErrorMessage(res: any) {
```
bjoerge: same as above re: `any`

src/index.ts
```diff
@@ -0,0 +1,18 @@
+import envMiddleware from './http/nodeMiddleware'
+import {defineHttpRequest} from './http/request'
+import {SanityClient} from './SanityClient'
+import type {ClientConfig} from './types'
+
+export * from './data/patch'
+export * from './data/transaction'
+export {ClientError, ServerError} from './http/errors'
+export * from './SanityClient'
```
bjoerge, line 9: Is there still a need for exporting the `SanityClient` class?

src/index.ts
```diff
@@ -0,0 +1,18 @@
+import envMiddleware from './http/nodeMiddleware'
+import {defineHttpRequest} from './http/request'
+import {SanityClient} from './SanityClient'
+import type {ClientConfig} from './types'
+
+export * from './data/patch'
+export * from './data/transaction'
```
bjoerge, line 7: Do we need these to be part of the public API?

bjoerge (approved):
This is fabulous!!! 💖 

Left a few minor comments/questions. Good to merge from my side.

src/generateHelpUrl.ts
```diff
@@ -1,5 +1,5 @@
 const BASE_URL = 'https://docs.sanity.io/help/'
```
kmelve:
```suggestion
const BASE_URL = 'https://www.sanity.io/help/'
```

The `docs.sanity.io` subdomain now just redirects to `sanity.io/docs`. The new format is `sanity.io/help`. For example: [https://www.sanity.io/help/studio-v2-vs-v3](https://www.sanity.io/help/studio-v2-vs-v3).

README.md
```diff
@@ -1,26 +1,85 @@
...
+
+Javascript client for Sanity. Works in [Node.js](https://nodejs.org/en/), [Bun], [Deno], [Edge Runtime] and [modern browsers].
+
+# Table of contents<!-- omit in toc -->
+
+- [@sanity/client](#sanityclient)
+  - [Requirements](#requirements)
+  - [Installation](#installation)
+  - [API](#api)
+    - [Creating a client instance](#creating-a-client-instance)
```
snorrees: Table of contents is awesome, and having detailed runtime instructions is great! (but) I find that big ones like this can look intimidating to first time users. 

Could we perhaps extract some of the finer details into `docs/<feature-details>.md` files and link them instead?  Or, consider putting install instructions and esm example on top?
Ie: put the common happy path on top so not to overwhelm anyone?

README.md
```diff
@@ -541,3 +816,292 @@ Semantic release will only release on configured branches, so it is safe to run
...
+
+### No longer shipping `ES5`<!-- omit in toc -->
+
+The target is changed to [modern browsers] that supports `ES6` `class`, `{...rest}` syntax and more. You may need to update your bundler to a recent major version. Or you could configure your bundler to transpile `@sanity/client`, and `get-it`, which is the engine that powers `@sanity/client` and uses the same output target.
+
+### Node.js `v12` no longer supported<!-- omit in toc -->
+
+Upgrade to the [LTS release, or one of the Maintenance releases](https://github.com/nodejs/release#release-schedule).
+
+### The `default` export is replaced with the named export `createClient`<!-- omit in toc -->
```
snorrees, line 893: This is the big one. I think it should be called out explicitly in the release notes, and not just link here. It will hit _everyone_ hard.

snorrees (approved):
This is such a massive effort! :head-exploding: :high-five:

Test coverage is so extensive, and types only adds safety so bug 👍 from me. 
Had a couple of thoughts, see other comments for those.

kmelve (approved):
Awesome work! This is really solid! 

I have pushed some updates to the client’s REAMDE.md. The only thing I wonder is if we should link to the compatibility table you made. Or do you prefer to keep that internal?

README.md
kmelve: Added it in a new commit. Please review @stipsan and @snorrees!

README.md
kmelve: This is the big one indeed. What is a good way to go about this? Add something like this? This also requires us to update all documentation and guides that have the client import. 

## Breaking change: 5.x

We have removed the default export and replaced it with a named one: 

```js
// Before Sanity Client v5
import SanityClient from '@sanity/client'

// After Sanity Client v5
```

.npmrc
stipsan: Good catch, looks like it's not 👍

src/http/errors.ts
stipsan: Good idea, an alias would let us remove `--quiet` from the linter command and still avoid all the noise

stipsan: There are no typings here because it wasn't defined in the original https://github.com/sanity-io/client/blob/main/sanityClient.d.ts#L884 file.
Fixing all of them is out of scope for this PR, I suggest doing that in follow ups

bjoerge: Makes perfect sense!

src/index.ts
stipsan: Yes, we don't expose an `httpRequest` option on `createClient`. But we do on the underlying classes. We're exposing `SanityClient` and `ObservableSanityClient` classes to allow other tooling to extend the client with a custom `httpTransport` without needing to reimplement large parts of the client.
Other exports are there to allow strict TS checks and not get complaints about a public interface using a type definition that isn't exported.

src/index.ts
stipsan: Yes, we have too much docs and references that do things like calling `new Transaction` and `new Patch`. The list over breaking changes is already quite long, and the `import SanityClient from '@sanity/client'` to `import {createClient} from '@sanity/client'` will create so much churn I feel it's best to not introduce further breaking changes. 😅

stipsan:
@kmelve feel free to link to it, I've cleaned up the repo and made it public: https://github.com/sanity-io/client-runtimes-compatibility
And the https://client-runtimes-dashboard.sanity.build/ still works

src/http/errors.ts
stipsan: @bjoerge I've replaced all the explicit `any` instances with the `FIXME` type, so we can follow up and incrementally fix them and improve the typings after v5 is released 🙇

README.md
stipsan: I like everything on one page and jumping between tabs is by far the most overwhelming thing I have to deal with every day. But I'm fine with being the outlier 🙇 
Looks good to me @kmelve!

README.md
stipsan: I've udpated the PR description and the commit body that'll be used. Semantic release will render something like this as release notes from it:

## [5.0.0](https://github.com/sanity-io/client/compare/v4.0.1...v5.0.0) (2023-01-13)


### ⚠ BREAKING CHANGES

We have removed the default export and replaced it with a named one:
  ```diff
  -import SanityClient from '@sanity/client'
  +import {createClient} from '@sanity/client'
  ```
[The migration guide outlines every breaking change and how to migrate your code](https://github.com/sanity-io/client/tree/modernize?rgh-link-date=2023-01-09T16%3A08%3A01Z#from-v4)

### Features

- full ESM support in modern runtimes and tooling (Bun, Deno, Edge, without breaking Node.js ESM and CJS compatibility)
- codebase rewritten in TypeScript, typings are generated and no longer manually maintained
- make `httpRequest` on `SanityClient` configurable, intended for libraries wishing to extend the client
- shipping modern syntax, reducing bundlesize

stipsan: We can enrich the release notes and changelog entry after semantic release have published it. But if we add more to the commit body we'll end up with Semantic Release arbitrarily cropping/mangling the output.

stipsan: Btw @kmelve  the new version will be `v5`. [`v4` is already released](https://github.com/sanity-io/client/releases/tag/v4.0.0), and don't have a migration guide as it didn't contain any API changes, it only added `pkg.exports` to add partial ESM support for runtimes that require it (like Cloudflare Workers and Edge Runtime on Next.js)

stipsan: >This is the big one indeed. What is a good way to go about this? 

Perhaps we could create a companion codemod for the imports change: https://github.com/facebook/jscodeshift @rexxars what do you think?

README.md
kmelve: Just realized that there is a complexity with this document also being the source of https://www.sanity.io/docs/js-client 🤔

README.md
rexxars: We could do that. Do you want me to write one?
Edit: Wrote one. See comment below.

rexxars:
@stipsan Here is a jscodeshift for the default import to named import:

```js
const importName = 'createClient'

v5ClientImport.parser = 'tsx'
module.exports = v5ClientImport

function v5ClientImport(fileInfo, api) {
  const j = api.jscodeshift
  const root = j(fileInfo.source)

  // Imports
  // from: import SanityClient from '@sanity/client'
  //   to: import {createClient as SanityClient} from '@sanity/client'
  //
  // from: import createClient, {MutationOptions} from '@sanity/client'
  //   to: import {createClient, MutationOptions} from '@sanity/client'
  root
    .find(api.jscodeshift.ImportDeclaration, (node) => node.source.value === '@sanity/client')
    .filter(
      (path) =>
        path.node.source.value === '@sanity/client' &&
        path.value.specifiers.some((specifier) => specifier.type === 'ImportDefaultSpecifier')
    )
    .forEach((path) => {
      const defaultSpecifier = path.value.specifiers.find(
        (specifier) => specifier.type === 'ImportDefaultSpecifier'
      )
      const namedImports = path.value.specifiers.filter(
        (specifier) => specifier !== defaultSpecifier
      )
      const oldName = defaultSpecifier.local.name

      j(path).replaceWith(
        j.importDeclaration(
          [j.importSpecifier(j.identifier(importName), j.identifier(oldName)), ...namedImports],
          j.literal('@sanity/client')
        )
      )
    })

  // Requires
  // from: const SanityClient = require('@sanity/client')
  //   to: const {createClient: SanityClient} = require('@sanity/client')
  root
    .find(j.VariableDeclarator, {
      id: {type: 'Identifier'},
      init: {callee: {name: 'require'}},
    })
    .filter((path) => path.value.init.arguments[0].value === '@sanity/client')
    .forEach((path) => {
      const oldName = path.value.id.name
      const prop = j.objectProperty(j.identifier(importName), j.identifier(oldName), false, true)

      // Avoid `const {createClient: createClient} = require('@sanity/client')`
      if (oldName === importName) {
        prop.shorthand = true
      }

      j(path).replaceWith(
        j.variableDeclarator(
          j.objectPattern([prop]),
          j.callExpression(j.identifier('require'), [j.stringLiteral('@sanity/client')])
        )
      )
    })

  return root.toSource()
}
```

Can be tested with the following command:

```
npx jscodeshift --extensions=js,jsx,ts,tsx -t pathToCodeshift.js targetFileOrDir
```

Or if you want to run it straight from a gist:

```
npx jscodeshift --extensions=js,jsx,ts,tsx -t https://gist.githubusercontent.com/rexxars/0d63e6e385adc78cbdc8b1cb4fd5cfcd/raw/f235b1b9dd520852c0bbf200c6174da5dafa9123/sanity-v5-codeshift.js targetFileOrDir
```

Successfully transforms the following examples:

CommonJS, retains variable name (rewritten to `const {createClient: createSanityClient} = ...`)
```js
const createSanityClient = require('@sanity/client')
const otherModule = require('@sanity/import')

createSanityClient({})
otherModule()
```

CommonJS, shorthand form because of same name as target being used (just adds curly brackets)
```js
const createClient = require('@sanity/client')
createClient({})
```

TypeScript/ESM: leaves unrelated imports alone, transforms to `import {createClient as SanityClient} from ...`
```ts
import SanityClient from '@sanity/client'
import otherModule from '@sanity/import'

const client = new SanityClient({})
```

TypeScript: Handles mixed default import + named import + type-only import (translates to `import {createClient as SanityClient, type MutationOptions} from ...`

```ts
import SanityClient, {type MutationOptions} from '@sanity/client'

const client = new SanityClient({})
const someOptions: MutationOptions = {dryRun: true}
client.create({_type: 'yes'}) 
```

We have some tooling for codemods in the _sanity_ CLI tool -  do we want to reuse that, or provide this as a standalone thing? Might be good enough to have people run the npx command?

stipsan:
@rexxars this is brilliant!
If moving it to the sanity cli still let's us run the codemod anywhere with just an npx one liner, without requiring the sanity monorepo package to be installed in the root package.json deps, and without requiring a sanity.confit.ts or sanity.cli.ts to exist... then that's where the codemod should be.
I expect us to add some improvements and fixes to it as people start using it in the wild.

Regarding the actual transformations, some feedback.
No matter if the source is
```ts
import SanityClient from "@sanity/client"
const client = new SanityClient(config)
```
or 
```ts
import createClient from "@sanity/client"
const client = createClient(config)
```
Then the end result should always be 
```ts
import {createClient} from "@sanity/client"
const client = createClient(config)
```

Application code should always use the names `createClient` export and call it like a typical function and never with `new`

The `import {SanityClient} from "@sanity"` named export is for a different use case, letting you change the httpRequest parameter that's how the first constructor argument while the client config is the second argument. This import is for library authors that build custom client tooling. And thus the codemod shouldn't transpile any v3 or v4 client code to creating instances of this class.

rexxars:
> Application code should always use the names `createClient` export and call it like a typical function and never with `new`

Hmm, that is new and breaking behavior - previously you could _either_ use it as a function _or_ call it with `new` (internally the function variant would instantiate a new class based on the config, if passed). If this behavior is changed, we need to explicitly call it out in the breaking changes, and we also need to update the codemod to alter any `new Client` calls to be the function variant.

In terms of the naming, it is a more complicated codemod to make, as you have to make sure `createClient` is not declared anywhere else, and update all the references to the function. I don't personally see any reason to go the extra mile on this - I might even prefer it to be alias to `createSanityClient` in my own code for readability.

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,35 @@
...
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  run:
+    name: 🤔
```
rexxars: I laughed, but maybe it should be slightly more descriptive? :p

README.md
```diff
@@ -1,26 +1,131 @@
...
+Sanity Client transpiles syntax for [modern browsers]. The JavaScript runtime must support ES6 features such as [class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes), [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters), [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and more. Most modern web frameworks, [browsers][modern browsers], and developer tooling supports ES6 today.
+
+For legacy ES5 environments you'll need to transpile `@sanity/client` and its `dependencies` with your own bundler, and have a global ES6-compliant `Promise` available. If your runtime environment doesn't provide a spec compliant `Promise` implementation, we recommend using [native-promise-only](https://www.npmjs.com/package/native-promise-only), [es6-promise](https://www.npmjs.com/package/es6-promise) or another [spec-compliant](https://promisesaplus.com/implementations) implementation. See [this article](https://www.sanity.io/help/js-client-promise-polyfill) for more information.
```
rexxars: We _could_ also tell people to keep using v4/v3 if they need to support older engines, I suppose?

README.md
```diff
@@ -31,13 +136,231 @@ const client = sanityClient({
...
+#### [TypeScript]
+
+```ts
+import {createClient, type ClientConfig} from '@sanity/client'
+
+const client: ClientConfig = {
+  projectId: 'your-project-id',
+  dataset: 'bikeshop',
+  apiVersion: '2021-03-25', // use current UTC date - see "specifying API version"!
+  token: 'sanity-auth-token', // or leave blank for unauthenticated usage
```
rexxars: Blank, undefined, or remove the property? I'm not sure what the code will do if you pass it a blank string, but I would expect passing undefined or not specifying the property to be the way to go.

README.md
```diff
@@ -31,13 +136,231 @@ const client = sanityClient({
...
+
+const client = createClient({
+  projectId: 'your-project-id',
+  dataset: 'bikeshop',
+  apiVersion: '2021-03-25', // use current UTC date - see "specifying API version"!
+  token: 'sanity-auth-token', // or leave blank for unauthenticated usage
+  useCdn: true, // `false` if you want to ensure fresh data
+})
+
+const data = await client.fetch(`count(*[])`)
```
rexxars: Some of these examples are all about instantiation of the client, which the header indicates, but others read more like usage examples - with fetch calls etc. Should we keep it consistent? (I think the query can be simplified to `count(*)` for what its worth)

README.md
```diff
@@ -31,13 +136,231 @@ const client = sanityClient({
...
+  const data = await client.fetch(`count(*[])`)
+  document.getElementById('results').innerText = `Number of documents: ${data}`
+</script>
+<div id="results"></div>
+```
+
+Or from anywhere using a dynamic `import()`:
+
+```js
+// You can run this snippet from your browwser DevTools console.
```
rexxars:
```suggestion
// You can run this snippet from your browser DevTools console.
```

README.md
```diff
@@ -155,7 +478,7 @@ client.createOrReplace(doc).then((res) => {
 })
 ```
 
-`client.createOrReplace(doc)`
```
rexxars, line 158: Are these trims intentional? They make markdown parsers output a linebreak instead of a paragraph, which makes the output a little more compact. Not a big deal for me, just wondering if intentional or not.

package.json
```diff
@@ -1,117 +1,128 @@
...
-    ],
-    "sourceMap": false
-  },
-  "prettier": {
-    "semi": false,
-    "printWidth": 100,
-    "bracketSpacing": false,
-    "singleQuote": true
+  "engines": {
+    "node": ">=14"
```
rexxars: Should we require node 14.18 or higher while we have the chance? I see a lot of dependencies (not necessarily ones we use in the client, but still) now using the `node:` prefix for imports, which is only supported on node 14.18 and higher.

src/auth/AuthClient.ts
```diff
@@ -0,0 +1,67 @@
+import {type Observable, lastValueFrom} from 'rxjs'
```
rexxars: Without having usage statistics here, I would be inclined to just remove this whole `auth` section of the client. Thoughts?

src/data/patch.ts
```diff
@@ -0,0 +1,357 @@
...
+   * DEPRECATED: Don't use.
+   * The operation is added to the current patch, ready to be commited by `commit()`
+   *
+   * @deprecated - Don't use.
+   * @param attrs - Attributes to replace
+   */
+  replace(attrs: AttributeSet): this {
+    validateObject('replace', attrs)
+    return this._set('set', {$: attrs})
+  }
```
rexxars: Given this has been deprecated for a while, should we remove it?

src/http/errors.ts
```diff
@@ -1,23 +1,43 @@
-import makeError from 'make-error'
+import {BaseError} from 'make-error'
```
rexxars: Does it still make sense to use `make-error`? I think we reached for it before ES6 - could we just go with `extends Error` these days?

test/encodeQueryString.test.ts
```diff
@@ -0,0 +1,30 @@
+import {expect, test} from 'vitest'
+
+import encode from '../src/data/encodeQueryString'
+
+test('can encode basic query without parameters', () => {
+  const query = 'gamedb.game[maxPlayers == 64]'
```
rexxars, line 6: Woof, these are some ancient queries 😅 
Might send a PR to replace these with some moe modern GROQ at some point ;)

rexxars (changes requested):
Amazing work! Left a few questions/comments inline.

src/http/errors.ts
stipsan: Great point, I checked with our browserslist query and we totally can! Will fix 👍

src/data/patch.ts
stipsan: We can, all the remaining deprecations are there because I couldn't find, or didn't know, how to explain the before/after migration for it. Do you have one for this?

src/auth/AuthClient.ts
stipsan: Sure, it's trivial to write a migration guide for those showing how to do the same things by calling `client.request` directly.
But, should we care about adding the auth typings to `client.request`?

If we do, it would mean that:
```ts
const loginProviders = await client.request({uri: '/auth/providers'})
// loginProviders is typed to `AuthProviderResponse`
```
And the same for `client.observable`.

If we don't, then the migration guide would look like this:
```ts
import {createclient, type AuthProviderResponse} from '@sanity/client'
const client = createClient()
const loginProviders = await client.request<AuthProviderResponse>({uri: '/auth/providers'})
```
Thoughts?

package.json
stipsan: In `get-it` [we've started running tests](https://github.com/sanity-io/get-it/blob/e26adb26281a974c4b8e67335f0418fd6b6aeadd/.github/workflows/ci.yml#L71) on the CI with `ls-engines` to catch that case. So if a dependency requires node 14.18 `ls-engines` will fail and output that `get-it` needs to change its `"node": ">=14"` `engines` condition to `"node": ">=14.18"`.
This PR adds the same check to `@sanity/client`.

Does it make sense to pre-emptively increase it now for `@sanity/client` so we don't have to make a `v6` release later? Or can it wait until either `ls-engines` says its time or the client itself wants to use `node:` prefixes?

README.md
stipsan: Did you make this change @kmelve? Or is it prettier being sneaky?

README.md
stipsan: Good point, v4 is stable and mature. What do you think @kmelve? It'll be annoying to keep this section up to date as we make use of other es6 features 😅

README.md
stipsan: How about
```suggestion
  token: 'sanity-auth-token', // needed if the `dataset` visibility is `private` or you want to query draft documents
```

README.md
stipsan: Fair point, I'll change them all to show `count(*)` 👍

.github/workflows/prettier.yml
stipsan: Is this better? 
```yml
    name: Can the code be prettier? 🤔
```

README.md
stipsan: I've updated these examples to just not use it as they're meant to be MVP snippets anyway.

So now only the first example mentions it, and perhaps that one should link to a docs page that explains a bit more what it's for? https://github.com/sanity-io/client/tree/modernize#quickstart

Especially since we have both viewer and editor tokens and we have 3 main use cases for it (from our users perspective):
- write tokens: mutate datasets, uploads assets
- viewer tokens for accessing private datasets
- viewer tokens for `@sanity/groq-store` (`next-sanity` and `@sanity/preview-kit`'s preview mode logic) to do `overlayDrafts`

stipsan: cc @kmelve

README.md
kmelve: I agree

README.md
kmelve: Good point. I kinda wanted to make it very clear in the quickstart that the client isn't only for querying documents, but also for updating them. 

Btw. have we thought about using `withCredentials` for preview stuff, or does that get messy because we also need server-side view access?

dburles:
Have you guys considered dropping CJS support entirely? That would free things up a great deal in how you can optimally publish the package across runtimes. See: https://jaydenseric.com/blog/optimal-javascript-module-design.

stipsan:
> Have you guys considered dropping CJS support entirely?

Dropping CJS wouldn't make it less work for us to support ESM. That part is completely automated with [`@sanity/pkg-utils`](https://github.com/sanity-io/client). For the last 12 months we've been migrating our own projects to run in ESM native mode and from that experience we know how painful it is to migrate when the npm ecosystem isn't ready for it.
It's best to wait until most projects are ESM :)

dburles:
What I mean is, right now you're publishing many different versions of the same thing to support multiple environments, along with an [ESM-wrapper](https://nodejs.org/dist/latest-v18.x/docs/api/packages.html#approach-1-use-an-es-module-wrapper) for Node, so Node isn't really consuming ESM. If this release was ESM only (with optimal module design), then ESM is all you would need to export. As you can imagine, that would be beneficial in many ways.

I'd consider going with that now, or at least set out a timeframe for going ESM only. In addition to the article I posted, here's some more reading https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c.

README.md
rexxars: > Btw. have we thought about using `withCredentials` for preview stuff, or does that get messy because we also need server-side view access?

`withCredentials` is becoming less of a thing you can depend on since more browsers are going cookieless

package.json
rexxars: I would say preemptively bump it to 14.18. It's a shame having to do a v6 release at an arbitrary point in time if one of the dependencies start using `node:`-prefixes

src/auth/AuthClient.ts
rexxars: 🤔 Good question. I think we should remove the typings from the client and instead export them from `@sanity/types` if they're not already there.

src/data/patch.ts
rexxars: The rationale is that it is unclear which properties this actually replaces in a document. What you usually want is either the `set()` patch (replacing the passed properties), or `createOrReplace()` (replacing the entire document)

rexxars: Before:
```ts
client
  .patch('tropic-hab')
  .replace({name: 'Tropical Habanero', ingredients: []})
  .commit()
```

After:
```ts
client.createOrReplace({
  _id: 'tropic-hab',
  _type: 'hotsauce',
  name: 'Tropical Habanero',
  ingredients: []
})
```

stipsan:
@dburles yes, you're right. We weighed the benefits of ESM-only against the cost it incurs to the ecosystem, we've seen how other library authors and consumers have reacted to [Sindre's decision](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c), we've migrated dozens if not hundreds of projects to use `"type": "module"` and know first hand how painful it can be to migrate legacy Jest, Webpack and TypeScript legacy setups to work with the new mode.

### How does our users benefit from migrating their project from CJS to ESM?
The truth is that for Node projects that work just fine on CJS the cost of migrating to ESM is really expensive, and what benefits are we giving them in return? [This thread gives valuable insights in how said projects respond to a library that drop CJS support](https://gist.github.com/joepie91/bca2fda868c1e8b2c2caf76af7dfcad3).
When CJS is letting them do exactly what they want, then what is ESM offering them? Any claim will have to be backed up with hard facts. Is ESM faster than CJS? They'll want you to show them numbers, and they better be good numbers. It's not enough to show numbers that prove that ESM is __technically__ faster. It'll need to be noticeably faster for anyone to care.

Then there's the projects that only supports CJS. Like [Electron](https://github.com/electron/electron/issues/21457#issuecomment-1321247244) and [React Native](https://github.com/facebook/metro/issues/886).
And the problem with TypeScript's Node resolution mode not knowing as well as most TypeScript projects that don't/can't use the new `Node16` resolution mode, and it throws `The requested module 'xyz' is a CommonJS module` errors even though the Node runtime won't throw and fully supports it. [I'm really happy that TS have realized their mistake with the `Node16` implementation and are addressing it in `v5` but that doesn't absolve us from taking TypeScript into consideration when we're considering dropping CJS.](https://github.com/microsoft/TypeScript/issues/50152)

### What's the cost of maintaining CJS support?

It's minimal, we already have world-class tooling that generates great CJS from our source ESM that has full TypeScript support and extremely wide ecosystem support ([here's our runtime support](https://client-runtimes-dashboard.sanity.build/), for bundlers we support the latest major of all of them). Then for Node specifically we just have to provide a wrapper to protect against the dual-package hazard scenario, that would otherwise have the potential for code like:
```ts
import {SanityClient} from '@sanity/client'

function logout(client) {
if(client instanceof SanityClient) { return client.auth.logout() }
}
```
to sometimes fail.
Since this is automated, what's left for us as maintainers to do to ensure CJS works? We add [one tiny smoke test](https://github.com/sanity-io/client/blob/modernize/runtimes/node/test.cjs) to the CI and basically never have to think about it again.
Again, the cost on the maintenance side is minimal, the entire cost is 1 smoke test, and [adding](https://github.com/sanity-io/client/blob/093648f2815a34937730c003c334b0630e51ded9/package.json#L32) [these](https://github.com/sanity-io/client/blob/093648f2815a34937730c003c334b0630e51ded9/package.json#L48) [lines](https://github.com/sanity-io/client/blob/093648f2815a34937730c003c334b0630e51ded9/package.json#L53) [to](https://github.com/sanity-io/client/blob/093648f2815a34937730c003c334b0630e51ded9/package.json#L39-L42) our package.json and boom we're shipping CJS.

What about the runtime cost? Like you mention, the wrapper is causing Node to always use the CJS runtime to load the client. Not letting you stay within the ESM runtime. This is true but what does this mean in practice?
Well, since CJS is far from slow this will either not make any difference at all to how fast the client loads, or [it'll load faster than if we let you load it using ESM](https://github.com/nodejs/node/issues/44186).

### What's our timeframe for dropping CJS and going ESM-only?

The conditions necessary for us to make this move are:
- TypeScript (likely v5) ships with a module resolver that follow `pkg.exports` semantics, knows that Node allows `@sanity/client` to `import` dependencies that are still CJS, and works with import path specifies that maintain compatibility with the rest of the tooling ecosystem (unlike Node16 that breaks with ESLint, Babel and practically every bundler).
- Electron and React Native are ready for ESM.
- We have a npm ecosystem that is ESM native, basically it should be enough to tell people to use the latest version of their linter, test runner etc instead of having to provide a long list over ["I'm having problems with ESM and xyz"](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#im-having-problems-with-esm-and-jest). As long as it's not enough for major projects like Jest that you have `"type": "module"` and valid `pkg.exports` to use ESM then the ecosystem simply isn't ready.
- The majority of projects that use `@sanity/client` are using `"type": "module"`, or there's a very clear and painless migration path for CJS folks to get on ESM.

The last point is most important. As long as shipping CJS is cheap for us to do and doesn't harm Node ESM performance, then it doesn't make sense for us to leave our CJS users out in the cold.

stipsan:
@dburles you're welcome to continue the conversation in a [new issue](https://github.com/sanity-io/client/issues/new), or if you have more questions ☺️ I'll be locking this conversation to collaborators so that we can shift focus back on landing the PR and avoid the potential for this to become a larger debate since there's a lot of people with strong opinions on the subject 🙌

rexxars:
I'm doing a check in the Sanity monorepo to see that the new client works as expected there (and migrating some code needed to make it work along the way). Will post any blockers here.

rexxars:
The image URL library depended on the `clientConfig` property, so needed a new check/code path to use the `config()` method instead: https://github.com/sanity-io/image-url/pull/51

rexxars:
Getting some errors on the `assets.upload` method when `assetType` is `'image' | 'file'` - it looks like adding a function overload that includes both makes it work, but maybe there is a better way?

```ts
  upload(
    assetType: 'file' | 'image',
    body: File | Blob | Buffer | NodeJS.ReadableStream,
    options?: UploadClientConfig
  ): Observable<
    HttpRequestEvent<{
      document: SanityAssetDocument | SanityImageAssetDocument
    }>
  >
```

README.md
stipsan: I've updated this section to read:
> [For legacy ES5 environments we recommend v4.](https://github.com/sanity-io/client/tree/v4.0.0#sanityclient)

README.md
stipsan: Yeah all our new templates with `live-preview` setup use a `token` due to the unreliability of `withCredentials`.

@kmelve I love the idea of using this line
```
// token: process.env.SANITY_SECRET_TOKEN // Only if you want to update content with the client
```
to bust the assumption that Studio is where you write, and `@sanity/client` is where you read.
So what do you think about expanding the pitch and adding a link?
```
// With a token you can query unpublished drafts, update content, upload images and files.
// https://www.sanity.io/docs/keeping-your-data-safe
// token: process.env.SANITY_SECRET_TOKEN
```
That's the best link I could find, but you get the idea :) 

Btw it looks like [these docs](https://www.sanity.io/help/js-client-usecdn-token) aren't updated with the the [v3 fix?](https://github.com/sanity-io/client/blob/371515b9c517c1849fcd2fc214872fb7fa7e7061/CHANGELOG.md?plain=1#L108)

stipsan:
> Hmm, that is new and breaking behavior - previously you could _either_ use it as a function _or_ call it with `new` (internally the function variant would instantiate a new class based on the config, if passed). If this behavior is changed, we need to explicitly call it out in the breaking changes, and we also need to update the codemod to alter any `new Client` calls to be the function variant.
> 
> In terms of the naming, it is a more complicated codemod to make, as you have to make sure `createClient` is not declared anywhere else, and update all the references to the function. I don't personally see any reason to go the extra mile on this - I might even prefer it to be alias to `createSanityClient` in my own code for readability.

Ok it sounds like it'll be a painful migration. The tricky bit here is that `httpRequest` is different in `index.ts` and `index.browser.ts` and is now the first constructor argument in `SanityClient`. In order to make the following codemod "just work":
```diff
-import SanityClient from '@sanity/client'
+import {SanityClient} from '@sanity/client'

const client = new SanityClient({projectId, ...etc})
```

Then the `index` files will have to re-export wrappers:

```ts
import envMiddleware from './http/nodeMiddleware'
import {defineHttpRequest} from './http/request'
import {SanityClient as BaseSanityClient} from './SanityClient'

const httpRequest = defineHttpRequest(envMiddleware)

export class SanityClient extends BaseSanityClient {
  constructor(config: ClientConfig) {
    super(httpRequest, config)
  }
}

export const createClient = (config: ClientConfig) => new BaseSanityClient(httpRequest, config)

```

Thoughts @rexxars? Yeah we should probably have the core classes change the `constructor` signature to match so it's consistent, but I'd like to hear your thoughts on the public API 🤔

src/auth/AuthClient.ts
stipsan: I couldn't get consensus for moving typings so for now I think we should just let them stay were they are and revisit :relaxed:

rexxars:
> Thoughts @rexxars? Yeah we should probably have the core classes change the `constructor` signature to match so it's consistent, but I'd like to hear your thoughts on the public API 🤔

Hmm, maybe we should just pull the plug on the constructor signature. I'll see if I can find time to update the codemod tomorrow 👍

rexxars:
@stipsan It looks like the `listen()` method is not accessible on the observable client. Intentional breaking change?

stipsan:
> @stipsan It looks like the `listen()` method is not accessible on the observable client. Intentional breaking change?

Unintentional! Fixing 🔨

rexxars (approved):
Have tested this extensively in the studio codebase - things seems to be working as expected! Will try to get a codemod ready for (constructor => factory method) tomorrow!

PR #92, opened 2023-01-19
feat: add support for abort controller

bjoerge:
**Status**:  Approved, waiting for #86 to land before merging

This adds support for passing an [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to all the request methods exposed by the client. It will also work for the APIs that returns an observable as well. Although less useful since observables natively supports cancellation, I can't see the problem in having support for it, but would love to hear your thoughts.

Note: I've copied the approach [ky](https://github.com/sindresorhus/ky/blob/740732c78aad97e9aec199e9871bdbf0ae29b805/source/errors/DOMException.ts) uses when it comes to creating an AbortError that works in Node < v18

Example usage:

```js
const abortController = new AbortController()
const request = client.fetch('*', {}, {signal: abortController.signal})

// this will abort the request and reject the `request` promise with an AbortError
abortController.abort()
```

### DIFF
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -447,6 +447,31 @@ describe('client', async () => {
       }
     })
 
+    test.skipIf(isEdge || typeof globalThis.AbortController === 'undefined')(
+      'can cancel request with an abort controller signal',
+      async () => {
+        expect.assertions(2)
+
+        nock(projectHost()).get(`/v1/data/query/foo?query=*`).delay(100).reply(200, {
+          ms: 123,
+          q: '*',
+          result: [],
+        })
+
+        const abortController = new AbortController()
+        const fetch = getClient().fetch('*', {}, {signal: abortController.signal})
+        await new Promise((resolve) => setTimeout(resolve, 10))
+
+        try {
+          abortController.abort()
+          await fetch
+        } catch (err: any) {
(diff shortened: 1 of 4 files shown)


stipsan (approved):
Let's do it let's go 🚀 would love it if you add an example for it in the readme with how you use it in an await fetch 🧘

test/client.test.ts
```diff
@@ -447,6 +447,27 @@ describe('client', async () => {
       }
     })
 
+    test.skipIf(false)('can cancel request with an abort controller signal', async () => {
```
stipsan:
```suggestion
    test.skipIf(isEdge && typeof AbortController === 'undefined')('can cancel request with an abort controller signal', async () => {
```
Doing this will ensure the test isn't run if the test run is in the vercel edge runtime, as it doesn't support `nock`. And it also skips older node environments that doesn't have `AbortController` so now the tests should pass

bjoerge:
> would love it if you add an example for it in the readme with how you use it in an await fetch 🧘

Great call! Added a couple of examples in [562b38b](https://github.com/sanity-io/client/pull/92/commits/562b38ba95314e182b26f6dfc3ba5221c96b519f)

rexxars (approved):
This is great! Thanks @bjoerge !

PR #148, opened 2023-03-09
feat: include mutation error items in error message

rexxars:
When you encounter mutation errors, the error message/staack from the client can be less than helpful:

```
ClientError: Mutation(s) failed with 1 error(s)
    at onResponse (/my/app/node_modules/@sanity/client/dist/index.cjs:78:13)
    at applyMiddleware (/my/app/node_modules/get-it/dist/index.cjs:60:15)
    ...
```

The response body does usually contain more information, but holds it in an `items` array. It is not obvious that this is available - you can technically get it at `error.details`, but it's typed as `any` and not documented. Ideally we'd introduce a new `MutationError` class with better typing, but that would be a breaking change (people may be relying on `instanceof ClientError` checks or similar), so will leave that for a later PR.

This PR provides a slight improvement on the situation, in a way that should be backwards compatible:
- It extracts the `description` property of the first 5 error items and includes them in the error message
- Includes the number of additional errors there were

So the resulting error might look like the following:

```
ClientError: Mutation(s) failed with 6 error(s):
- Malformed document ID: "#some_invalid-id!"
- Malformed document ID: "@ruby_bird@"
- Malformed document ID: "!cant_contain_that"
- Malformed document ID: "what$about!this?"
- Malformed document ID: "%so_many_percent%"
...and 1 more
    at onResponse (/my/app/node_modules/@sanity/client/dist/index.cjs:78:13)
    at applyMiddleware (/my/app/node_modules/get-it/dist/index.cjs:60:15)
    ...
```

### DIFF
--- a/src/http/errors.ts
+++ b/src/http/errors.ts
@@ -1,5 +1,7 @@
 import type {Any, ErrorProps, MutationError} from '../types'
 
+const MAX_ITEMS_IN_ERROR_MESSAGE = 5
+
 /** @public */
 export class ClientError extends Error {
   response: ErrorProps['response']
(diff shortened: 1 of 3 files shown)


stipsan (approved):
Nice! 👍

src/http/errors.ts
```diff
@@ -44,6 +44,22 @@ function extractErrorProps(res: Any): ErrorProps {
     return props
   }
 
+  // Mutation errors (specifically)
+  if (isMutationError(body)) {
+    const allItems = body.error.items || []
+    const items = allItems
+      .slice(0, 5)
```
bjoerge: If we want to change the number of errors shown by default, it's easy to miss one of these `5`'s and get an inaccurate message. Put it in a constant instead?

bjoerge (approved):
This is great! 💯 

Just a minor nit, otherwise LGTM

src/http/errors.ts
rexxars: Great point! Have changed to a constant.

PR #149, opened 2023-03-09
fix: correct document ID validation

rexxars:
The current document ID validation is a little too lenient:

- Allows IDs that start with a dash
- Allows double-dots (`some..path`)
- Allows long IDs (max length is 128)

This PR fixes these issues.


stipsan (approved):
Good fix 😌

Issue #177, opened 2023-03-23
Adding Sanity client breaks new Hydrogen storefront


LukasMurdock:
**Steps to reproduce**

1. Create a new Hydrogen app: [https://shopify.dev/docs/custom-storefronts/hydrogen/getting-started/quickstart](https://shopify.dev/docs/custom-storefronts/hydrogen/getting-started/quickstart)
2. Choose `hello world` template
3. Choose `TypeScript`
4. Install dependencies
5. Install Sanity client
6. Run `npm run dev` and see it work
7. Create new Sanity client in the loader function in `root.tsx`
8. Run `npm run dev` and see it break

**Expected behavior**

Adding Sanity client does not break anything, and works.

**Actual behavior**

Adding Sanity client throws `TypeError: globalThis.XMLHttpRequest is not a constructor` and does not work.

frehner:
There appears to be some devs that have found downgrading to `5.2.2` has fixed things for them.

dylanmcgowan:
> There appears to be some devs that have found downgrading to `5.2.2` has fixed things for them.

Downgrading did not work for me :(

christiandunbar:
@LukasMurdock @dylanmcgowan 

The following packages and versions for me locally

```
"@remix-run/react": "1.14.0"
"@sanity/block-content-to-react": "3.0.0"
"@sanity/client": "5.3.0"
"@sanity/types": "2.35.0"
"@shopify/cli": "3.44.1"
"@shopify/cli-hydrogen": "4.0.9"
"@shopify/hydrogen": "2023.1.6"
"@shopify/remix-oxygen": "1.0.4"
```

It however does not work when deploying to Shopify Oxygen when using the automatically generated Github action file as it uses yarn which causes the same `TypeError: globalThis.XMLHttpRequest is not a constructor` error.

I got the deployment to work when changing from yarn to npm.

![image](https://user-images.githubusercontent.com/11332936/227660039-8204790a-067a-4aa3-b4b4-f10f0793b928.png)

dcodrin:
This is probably the change that introduced the issue https://github.com/sanity-io/client/pull/169.

gmisa:
Tried downgrading to **5.2.2** and replaced yarn with **npm** but still the same issue.  `TypeError: globalThis.XMLHttpRequest is not a constructor`.

dcodrin:
@gmisa you have to make sure to regenerate your yarn.lock file.
[...]

PR #181, opened 2023-03-28
fix(listen): lazy init event source

stipsan:
Reduces bundle size for environments which doesn't use `client.listen`.
Also helps cases like #177.

Can be tested using: `npm i --save-exact @sanity/client@lazy-event-source`.

Build output diff: https://app.renovatebot.com/package-diff?name=@sanity/client&from=5.4.0&to=5.4.1-lazy-event-source.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sanity/client",
-  "version": "5.4.0",
+  "version": "5.4.1-lazy-event-source.0",
   "description": "Client for retrieving, creating and patching data from Sanity.io",
   "keywords": [
     "sanity",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sanity/client",
-  "version": "5.4.0",
+  "version": "5.4.1-lazy-event-source.0",
```
bjoerge, line 3: does this need to change before merge?

bjoerge (approved):
This is great, lgtm!

package.json
stipsan: Nope, semantic release will handle it 😌

PR #186, opened 2023-04-03
fix(mutate): serialize clientless patches correctly

rexxars:
This PR fixes an issue where passing a clientless patch (`new Patch('doc-id')`) will result in the mutation getting incorrectly serialized when passed to `mutate()`.

Actually serialized body:

```js
{mutations: [{id: 'doc-id', set: {foo: 'bar'}]}
```

Expected body:

```js
{mutations: [{patch: {id: 'doc-id', set: {foo: 'bar'}]}}
```


stipsan (approved):
Good catch!!

PR #187, opened 2023-04-03
fix(mutate): reflect support for providing transaction id in typings

rexxars:
The `mutate()` method takes a `transactionId` option and passes it along to the request body - but this was not reflected in the typings. This PR ensures the typings are up to date, and removes the use of a now unnecessary `as any` cast.


stipsan (approved):
LGTM 👍

PR #191, opened 2023-04-04
fix!: make `useCdn` use `true` by default

stipsan:
# Why `true` is better
Seeing unexpected stale content is a far better problem to have than a surprise overage bill 😅 
[This is a follow up to this discussion.](https://sanity-io.slack.com/archives/C043ZRB9E4S/p1680101505217759)

Also worth noting that one of the historical reasons why `useCdn: false` have been the default were the limitation the client had prior to `v3`: https://www.sanity.io/help/js-client-cdn-configuration#:~:text=subscribing%20to%20changes).-,Gotcha,-Prior%20to%20v3.0.0
Now that the client have supported using `token` with `useCdn: true` for quite some time it makes sense to change the default.

# This will be a breaking release
The release note will highlight this as:
BREAKING CHANGE: if you want to keep the previous behavior, add `useCdn: false` to your configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ import {createClient} from '@sanity/client'
 export const client = createClient({
   projectId: 'your-project-id',
   dataset: 'your-dataset-name',
-  useCdn: false, // set to `true` to fetch from edge cache
+  useCdn: true, // set to `false` to bypass the edge cache
   apiVersion: '2022-01-12', // use current date (YYYY-MM-DD) to target the latest API version
   // token: process.env.SANITY_SECRET_TOKEN // Only if you want to update content with the client
 })
@@ -93,7 +93,9 @@ export async function updateDocumentTitle(_id, title) {
   - [Set client configuration](#set-client-configuration)
 - [Release new version](#release-new-version)
 - [License](#license)
-- [From `v4`](#from-v4)
+- [Migrate](#migrate)
+  - [From `v5`](#from-v5)
+  - [From `v4`](#from-v4)
 
 ## Requirements
 
@@ -119,7 +121,7 @@ pnpm install @sanity/client
 
 `const client = createClient(options)`
 
-Initializes a new Sanity Client. Required options are `projectId`, `dataset`, and `apiVersion`. Setting a value for `useCdn` is encouraged. Typically you want to have it as `false` in development to always fetch the freshest content and `true` in production environments so that content is fetched from the distributed cache. [You can learn more about the API CDN here][api-cdn].
+Initializes a new Sanity Client. Required options are `projectId`, `dataset`, and `apiVersion`. [We encourage setting `useCdn` to either `true` or `false`.](https://www.sanity.io/help/js-client-cdn-configuration) The default is `true`. If you're not sure which option to choose we recommend starting with `true` and revise later if you find that you require uncached content. [Our awesome Slack community can help guide you on how to avoid stale data tailored to your tech stack and architecture.](https://slack.sanity.io/)
 
 #### [ESM](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/)
 
@@ -129,7 +131,7 @@ import {createClient} from '@sanity/client'
 const client = createClient({
   projectId: 'your-project-id',
   dataset: 'your-dataset-name',
-  useCdn: false, // set to `true` to fetch from edge cache
+  useCdn: true, // set to `false` to bypass the edge cache
   apiVersion: '2022-01-12', // use current date (YYYY-MM-DD) to target the latest API version
 })
 
@@ -145,7 +147,7 @@ const {createClient} = require('@sanity/client')
 const client = createClient({
   projectId: 'your-project-id',
   dataset: 'your-dataset-name',
-  useCdn: false, // set to `true` to fetch from edge cache
+  useCdn: true, // set to `false` to bypass the edge cache
   apiVersion: '2022-01-12', // use current date (YYYY-MM-DD) to target the latest API version
 })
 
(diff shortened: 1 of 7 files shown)


kmelve:
This is a neat way of preventing some of the unintended cases of hitting the API limits.

The semver change seems a bit tricky, but then again, everyone who have been using this with undefined should have seen the warning. It seems unlikely that someone are relying on the undefined behavior for business critical functionality. 

That being said, what's unclear to me is how we can make sure that the developer experience for folks that use the client to fetch content for static rendering isn't getting more confusing, as well as typical onboarding scenarios, and local development feedback loops. 

## New users and onboarding 
How will this affect developers new to the platform who are asked to publish a document and see their change in the frontend? Maybe the cache invalidation is quick enough that it's fine? Either way, we should make sure to make a note of this property in new onboarding materials. 

## Documentation
This release could incentivize writing better docs on how our caching actually works, including invalidation. I see the "why aren't I'm seeing my published content" quite often. 

It would also be very helpful if we provided guidance on how to set this property so it works with typical static rendering where builds or frontend revalidation is triggered by a webhook. This also affect content creators who are confused by not seeing their published content. I get that it's slightly outside of this change specifically, but it does bring it into focus.

stipsan:
To solve the general problem of stale content we could, when publishing a document, show an intermediary "yellow" state: Draft -> Pending/Propagating/Publishing -> Published. 
This state can be disabled in the studio config, it's enabled by default. A mouseover tells you if it's fully available in the API CDN yet or how long until it is (we have metrics on the timing of this).

For the static site building I think it needs better docs indeed that is tailored specifically for that. There's a lot of moving parts, ISR or no, webhooks or no, etc.

README.md
```diff
@@ -119,7 +119,7 @@ pnpm install @sanity/client
 
 `const client = createClient(options)`
 
-Initializes a new Sanity Client. Required options are `projectId`, `dataset`, and `apiVersion`. Setting a value for `useCdn` is encouraged. Typically you want to have it as `false` in development to always fetch the freshest content and `true` in production environments so that content is fetched from the distributed cache. [You can learn more about the API CDN here][api-cdn].
+Initializes a new Sanity Client. Required options are `projectId`, `dataset`, and `apiVersion`. [We encourage setting `useCdn` to either `true` or `false`.](https://www.sanity.io/help/js-client-cdn-configuration) It's `true` by default. If you're not sure which option to use then stick to `true` and revisit later should stale data prove to cause problems. [Our awesome Slack community can help guide you on how to avoid stale data tailored to your tech stack and architecture.](https://slack.sanity.io/)
```
runeb: @stipsan Suggest changing
> If you're not sure which option to use then stick to `true` and revisit later should stale data prove to cause problems

to

> If you're not sure which option to choose we recommend starting with `true` and revise later if you find that you require uncached content

src/warnings.ts
```diff
@@ -7,8 +7,8 @@ const createWarningPrinter = (message: string[]) =>
   once((...args: Any[]) => console.warn(message.join(' '), ...args))
 
 export const printCdnWarning = createWarningPrinter([
-  'You are not using the Sanity CDN. That means your data is always fresh, but the CDN is faster and',
-  `cheaper. Think about it! For more info, see ${generateHelpUrl('js-client-cdn-configuration')} `,
+  `You're using the Sanity API CDN. Its edge caching makes it faster and cheaper than the live API. `,
```
runeb, line 10: Suggest changing this wording to something like

> You're using the Sanity API CDN which has the fastest and most scalable content delivery. You can opt out if you require uncached content.

README.md
runeb:
```suggestion
Initializes a new Sanity Client. Required options are `projectId`, `dataset`, and `apiVersion`. [We encourage setting `useCdn` to either `true` or `false`.](https://www.sanity.io/help/js-client-cdn-configuration) The default is `true`. If you're not sure which option to choose we recommend starting with `true` and revise later if you find that you require uncached content. [Our awesome Slack community can help guide you on how to avoid stale data tailored to your tech stack and architecture.](https://slack.sanity.io/)
```

src/warnings.ts
```diff
@@ -7,8 +7,8 @@ const createWarningPrinter = (message: string[]) =>
...
+  `For more info, see ${generateHelpUrl('js-client-cdn-configuration')} `,
   'To hide this warning, please set the `useCdn` option to either `true` or `false` when creating',
   'the client.',
```
runeb:
```suggestion
  `Since you haven't set a value for /`useCdn/`, we will deliver content using our`,
  `global, edge-cached API-CDN. If you wish to have content delivered faster, set`,
  `/`useCdn: false`/ to use the Live API. Note: You may incur higher costs using the live API.`,
```

test/warnings.test.ts
```diff
@@ -15,7 +15,7 @@ describe('Client config warnings', async () => {
   test('warns if useCdn is not given', () => {
     createClient({projectId: 'abc123', apiVersion: '1'})
     expect(warn).toHaveBeenCalledWith(
-      'You are not using the Sanity CDN. That means your data is always fresh, but the CDN is faster and cheaper. Think about it! For more info, see https://www.sanity.io/help/js-client-cdn-configuration  To hide this warning, please set the `useCdn` option to either `true` or `false` when creating the client.'
+      "You're using the Sanity API CDN. Its edge caching makes it faster and cheaper than the live API.  For more info, see https://www.sanity.io/help/js-client-cdn-configuration  To hide this warning, please set the `useCdn` option to either `true` or `false` when creating the client."
```
runeb:
```suggestion
      "Since you have not set a value for useCDN, we will deliver content using our global, edge-cached API-CDN. If you wish to have content delivered faster, set useCDN: false to use the Live API. Note: You may incur higher costs using the live API."
```

runeb (dismissed):

runeb:
Added suggested changes from @tarungangwani

rexxars (approved):
While the change in and of itself looks fine, I don't agree that this shouldn't be considered a breaking change. People may have just ignored the warnings on app startup, for instance. I'd vote to bump the major version.

PR #199, opened 2023-04-18
feat!: Add automatic retrying of 429/502/503

judofyr:
This is a bit of a bigger, and technically breaking, change which introduces automatic retrying of 429 errors (i.e. rate limited). I'm opening it here to start a discussion about how to progress this forward. I believe this would be very useful for the majority of the users of the client and should be enabled by default.

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -1,5 +1,5 @@
 import {getIt, type Middlewares} from 'get-it'
-import {jsonRequest, jsonResponse, observable, progress} from 'get-it/middleware'
+import {jsonRequest, jsonResponse, observable, progress, retry} from 'get-it/middleware'
 import {Observable} from 'rxjs'
 
 import type {Any, HttpRequest, RequestOptions} from '../types'
@@ -27,8 +27,22 @@ const printWarnings = {
 }
 
 /** @internal */
-export function defineHttpRequest(envMiddleware: Middlewares): HttpRequest {
+export function defineHttpRequest(
+  envMiddleware: Middlewares,
+  {
+    maxRetries = 5,
+    retryDelay,
+  }: {maxRetries?: number; retryDelay?: (attemptNumber: number) => number}
+): HttpRequest {
   const request = getIt([
+    maxRetries > 0
+      ? retry({
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          retryDelay: retryDelay as any, // This option is typed incorrectly in get-it.
+          maxRetries,
+          shouldRetry,
+        })
+      : {},
     ...envMiddleware,
     printWarnings,
     jsonRequest(),
@@ -45,4 +59,30 @@ export function defineHttpRequest(envMiddleware: Middlewares): HttpRequest {
   httpRequest.defaultRequester = request
 
   return httpRequest
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function shouldRetry(err: any, attempt: number, options: any) {
+  // By default `retry.shouldRetry` doesn't retry on server errors so we add our own logic.
+
+  const isSafe = options.method === 'GET' || options.method === 'HEAD'
+  const isQuery = options.uri.startsWith('/data/query')
+  const isRetriableResponse =
+    err.response &&
+    (err.response.statusCode === 429 ||
+      err.response.statusCode === 502 ||
+      err.response.statusCode === 503)
+
+  // We retry the following errors:
+  // - 429 means that the request was rate limited. It's a bit difficult
+  //   to know exactly how long it makes sense to wait and/or how many
+  //   attempts we should retry, but the backoff should alleviate the
+  //   additional load.
+  // - 502/503 can occur when certain components struggle to talk to their
+  //   upstream dependencies. This is most likely a temporary problem
+  //   and retrying makes sense.
+
+  if ((isSafe || isQuery) && isRetriableResponse) return true
+
+  return retry.shouldRetry(err, attempt, options)
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,12 +11,19 @@ export * from './SanityClient'
 export * from './types'
 
 // Set the http client to use for requests, and its environment specific middleware
-const httpRequest = defineHttpRequest(envMiddleware)
+const httpRequest = defineHttpRequest(envMiddleware, {})
 /** @public */
 export const requester = httpRequest.defaultRequester
 
 /** @public */
-export const createClient = (config: ClientConfig) => new SanityClient(httpRequest, config)
+export const createClient = (config: ClientConfig) =>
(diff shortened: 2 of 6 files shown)


stipsan (commented):
could you change the PR title to: feat!: add automatic retry on 429

Just so that semantic release bumps the right version and generates appropriate release notes. We can do this major release together with the useCdn one so we don't have two breaking releases in short succession

judofyr:
> could you change the PR title to: feat!: add automatic retry on 429

It's enough with the PR title? Or should the commit also have the `!`?

stipsan:
Commit title as well

judofyr:
- Changed the commit message to `feat!`
- Added retrying of 502 and 503 as well.
- Added more comment explaining why we retry.

src/http/request.ts
```diff
@@ -27,8 +27,39 @@ const printWarnings = {
...
+      shouldRetry(err, attempt, options) {
+        // By default `retry.shouldRetry` doesn't retry on server errors so we add our own logic.
+
+        const isSafe = options.method === 'GET' || options.method === 'HEAD'
+        const isQuery = options.uri.startsWith('/data/query')
+        const shouldRetry =
+          err.response &&
+          (err.response.statusCode === 429 ||
+            err.response.statusCode === 502 ||
+            err.response.statusCode === 503)
```
rexxars: There is a case to be made for also retrying other non-response errors. For instance, temporary socket errors, DNS resolve failures etc. Normally these wouldn't be retried if the request is a POST-request - but we could retry POST _queries_ at least.

Additionally, DNS lookup failures would prevent a request from even going out, so we could also retry those if the URL is still `*api.sanity.io`/`*apicdn.sanity.io`, since those _should_ always resolve. I've seen multiple reports of DNS lookup errors being reported, and I assume it is a temporary resolve failure on the client side causing it since the alternative would show pretty rapidly on our charts (and reported issues).

Here are some common error codes: https://nodejs.org/api/errors.html#common-system-errors

src/index.ts
```diff
@@ -11,12 +11,19 @@ export * from './SanityClient'
 export * from './types'
 
 // Set the http client to use for requests, and its environment specific middleware
-const httpRequest = defineHttpRequest(envMiddleware)
+const httpRequest = defineHttpRequest(envMiddleware, {maxRetries: 0})
```
rexxars: What is the rationale/intended behavior of `maxRetries: 0` here?
[This line](https://github.com/sanity-io/get-it/blob/main/src/middleware/retry/shared-retry.ts#L8) in get-it leads me to believe that it would result in the _actual_ max retries being set to 5, which may or may not be what you want, depending on the intention.

src/http/request.ts
```diff
@@ -27,8 +27,39 @@ const printWarnings = {
...
-export function defineHttpRequest(envMiddleware: Middlewares): HttpRequest {
+export function defineHttpRequest(
+  envMiddleware: Middlewares,
+  {maxRetries, retryDelay}: {maxRetries: number; retryDelay?: (attemptNumber: number) => number}
+): HttpRequest {
   const request = getIt([
+    retry({
+      retryDelay: retryDelay as () => number, // This option is typed incorrectly in get-it.
+      maxRetries,
+      shouldRetry(err, attempt, options) {
```
rexxars: Could we export this for people that want to reuse it in their custom `shouldRetry` functions?

rexxars (changes requested):
- This looks great - will be really nice to get those temporary errors retried. I think this is unlikely to "solve" rate limiting (with such low first-retry timeouts), but also probably won't make it worse. I Have A Branch™ that implements actual rate-limiting that I'd like to finish up at some point, but we can leave that for a later time.

- Maybe we should provide a way to disable the retrying? Passing `{maxRetries: 0}` will actually end up setting it to `5` currently (unfortunate choice, I guess we might want to fix this in get-it).

- I would prefer if we also added an explicit `BREAKING CHANGE:` explanation to the commit(s), in order for explain it in a little more detail for automatic release notes. For instance:
```
BREAKING CHANGE: Client will now automatically retry all GET/HEAD requests as
well as queries if the server responds with a 429, 502 or 503 status code - as
well as on socket/dns errors. Previously, the client would immediately throw
an error. If you have application-level retry code, you should either disable
the retrying in the client by passing `{maxRetries: 0}`, or remove the custom
retry code and potentially alter the `shouldRetry` and `maxRetries` options to
match your wanted behavior.
```

src/http/request.ts
judofyr: We're already using `is-retry-allowed` in get-it which I believe should support this, but it's not applied to POST. Maybe the "proper" way to support this is to support `options.isSafe = true` (or maybe `options.isIdempotent = true` since we can actually also retry these?), and then have the default `shouldRetry` check for this instead?

```js
function shouldRetry(…) {
  const isSafe = options.isSafe || options.method === 'GET' || options.method === 'HEAD' // remain backwards compatible
  if (!isSafe) return false
}
```

src/index.ts
judofyr: The _intention_ was that the requester would get no retries in order to make it "less" of a breaking change. I'm not sure if this makes sense though. It's maybe more consistent to default to 5 there as well?

src/http/request.ts
judofyr: Currently `shouldRetry` _isn't_ exposed as a public options. The intention was that this is something which shouldn't be configured by an end-user as this is determined by what the Sanity backend supports (and hence should be "hard coded" in the client).

judofyr:
Okay, I've made the following changes:

- Made it possible to disable the retrying logic by specifying `maxRetries: 0`.
- Updated commit message.
- Removed the `retry` that we used in the Node middleware. (Wasn't aware of this!)
- Changed the exported `requester` to have the same retry logic.

I'd prefer to merge this now and then rather implement "retry also on DNS errors for POST requests" in a later (patch) release as I think this is best solved by adding support for it in `get-it`.

rexxars (approved):

src/index.ts
```diff
@@ -11,12 +11,19 @@ export * from './SanityClient'
 export * from './types'
 
 // Set the http client to use for requests, and its environment specific middleware
-const httpRequest = defineHttpRequest(envMiddleware)
+const httpRequest = defineHttpRequest(envMiddleware, {})
 /** @public */
 export const requester = httpRequest.defaultRequester
```
stipsan, line 16: Out of scope for this PR, but it does seem like we should export `defineHttpRequest` instead of `requester`. I'm not sure what use there is for `requester` anymore. What do you think @rexxars?

stipsan (approved):
All good!

src/index.ts
rexxars: There are a few usages internally, but I'm sure they could be done differently: https://github.com/search?q=org%3Asanity-io+requester&type=code

PR #205, opened 2023-04-26
fix: undeprecate `request()`

rexxars:
The reason it was marked as deprecated was to steer users away from it in favor or using their own request library, since we didn't want the maintenance burden of having to keep request options/behavior compatible between releases. However, after seeing a number of real-world use cases where users want to reuse the client to do requests against Sanity API endpoints that do not have their own built-in method (scheduled publishing, webhook APIs etc), and the complexity of figuring out the authentication part (cookie? auth token?), we decided it is a reasonable ask to allow using the method.

The `dataRequest()` is more of an internal convenience method on top of `request()`, and should still be considered deprecated, since the same tasks can easily be achieved using `request()`. However, I did clarify it a bit in the deprecated note.

### DIFF
--- a/src/SanityClient.ts
+++ b/src/SanityClient.ts
@@ -1210,22 +1210,25 @@ export class SanityClient {
   }
 
   /**
-   * DEPRECATED: Perform an HTTP request against the Sanity API
+   * Perform a request against the Sanity API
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
Great call, lgtm!

stipsan (approved):
Sure, makes sense to me 😌

PR #218, opened 2023-05-04
feat: add `scope` parameter

stipsan:
Can be tested using `npm i @sanity/client@scope`. Requires `apiVersion: 'X'` and shouldn't be used in production!


stipsan:
Replaced by #224

PR #222, opened 2023-05-09
Update CHANGELOG.md, fix link to doc

stoivo:
The link didn't work from viewing the changelog from GitHub so I think it was a mistake.


stipsan (approved):
Good catch, thanks!

PR #232, opened 2023-05-23
fix: update `RawQueryResponse` response type properties

dankreiger:
Hello everyone 👋 

I've noticed a type definition issue while utilizing the `fetch` method on `@sanity/client`. The response returned from the API seems to have a slight discrepancy from what's defined in the `RawQueryResponse` type.

On the `RawQueryResponse`:

- Rather than receiving `q`, I'm receiving `query`.
- The `resultSourceMap` is optional and doesn't appear as a property unless it's part of the input parameters.

CleanShot 2023-05-23 at 11 31 53

The URL structure produced by the `fetch` method  adheres to the standard format from the Sanity API:

```
https://[projectId].api.sanity.io/[apiVersion]/data/query/[dataset]?query=[groqString]
```

I've also attempted to test this with varying API versions, and both with and without the CDN.

***

Could I possibly be doing something incorrectly or using the wrong type? If that's the case, feel free to close this discussion.

Thank you very much for maintaining this package!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -489,10 +489,10 @@ export type UnfilteredResponseQueryOptions = RequestOptions & {
 
 /** @public */
 export interface RawQueryResponse<R> {
-  q: string
+  query: string
   ms: number
   result: R
-  resultSourceMap: ContentSourceMap
+  resultSourceMap?: ContentSourceMap
 }
 
 /** @internal */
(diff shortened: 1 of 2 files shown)


stipsan (approved):
Thanks, great find! 🙌

PR #249, opened 2023-06-28
feat: support `fetch` for Next.js `app-router`

stipsan:
Testable using:
```bash
npm i @sanity/client@fetch --save-exact
```

It allows opt-in to using native `fetch` in the `nodejs` runtime in Next.js:
```ts

const client = createClient({
  ...config,
  // use `fetch`, opts-in to automatic fetch deduping and other benefits in app-router
  fetch: true,
})
```

And in both `nodejs`, and now `edge`, you can set certain useful options like `cache` and `next` on the request level:
```ts
const data = await client.fetch(query, params, {
  cache: 'force-cache',
  next: {revalidate: 30, tags: ['page']}
})
```

This allows us to get rid of workarounds such as: https://github.com/sanity-io/next-sanity#app-router-react-server-components-and-caching
And the ability to use features that are only available to [data fetching that use `fetch`
 under the hood.](https://vercel.com/docs/infrastructure/data-cache#tag-based-revalidation)


rexxars (approved):

gksander (approved):
Awesome! Tried this on my end in a production app running Next.js 13 (using fetch cache tags etc etc), and it appears to Just Work ™️

indeox:
Tested this as well, and works beautifully. Very much looking forward to when it gets merged! 🎉  It will solve so many issues

PR #252, opened 2023-06-29
fix: validate `perspective` config

stipsan:
@snorrees do you know how we can widen the `perspective` type to allow strings, without losing IDE autocomplete?

### DIFF
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -161,6 +161,22 @@ describe('client', async () => {
       expect(() => createClient({projectId: 'abc123', logger: console})).toThrow(/logger/)
     })
 
+    test('throws on invalid perspective', () => {
+      expect(() => createClient({projectId: 'abc123', perspective: 'published'})).not.toThrow(
+        /Invalid API perspective/
+      )
+      expect(() => createClient({projectId: 'abc123', perspective: 'previewDrafts'})).not.toThrow(
+        /Invalid API perspective/
+      )
+      expect(() => createClient({projectId: 'abc123', perspective: 'raw'})).not.toThrow(
+        /Invalid API perspective/
+      )
+      // @ts-expect-error -- we want to test that it throws an error
+      expect(() => createClient({projectId: 'abc123', perspective: 'preview drafts'})).toThrow(
+        /Invalid API perspective/
+      )
+    })
+
     test('throws on invalid project ids', () => {
       expect(() => createClient({projectId: '*foo*'})).toThrow(/projectId.*?can only contain/i)
     })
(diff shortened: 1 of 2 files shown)


snorrees:
> @snorrees do you know how we can widen the `perspective` type to allow strings, without losing IDE autocomplete?

Only for IntelliJ, not VsCode unfortunatly (MyType | string).  Whats in the PR works for IntelliJ. 

Could try function overloading and see if it changes anything (guessing here)?

snorrees:
> > @snorrees do you know how we can widen the `perspective` type to allow strings, without losing IDE autocomplete?
> 
> Only for IntelliJ, not VsCode unfortunatly (MyType | string). Whats in the PR works for IntelliJ.
> 
> Could try function overloading and see if it changes anything (guessing here)?

Try this. It was suggested recently for core apis but there it lead to cascading inferrence issues. You probably wont have that problem here.

https://github.com/microsoft/TypeScript/issues/29729#issuecomment-567871939

snorrees (commented):
Seems like autocomplete fix is in typescript 5.2 iteration plan

stipsan:
Unable to get it to work, but if 5.2 fixes it maybe that's ok for now @snorrees?

src/types.ts
```diff
@@ -29,7 +29,7 @@ export interface RequestOptions {
 }
 
 /** @public */
-export type ClientPerspective = 'previewDrafts' | 'published' | 'raw'
+export type ClientPerspective = ('previewDrafts' | 'published' | 'raw') | string
```
snorrees: Is adding string a future-proofing? Cause according to your validator, you dont accept string, so why add the string? Isnt it better to get the compiler to give early error? (keep the validator tho, defensive programming ect)

test/client.test.ts
```diff
@@ -161,6 +161,21 @@ describe('client', async () => {
       expect(() => createClient({projectId: 'abc123', logger: console})).toThrow(/logger/)
     })
 
+    test('throws on invalid perspective', () => {
+      expect(() => createClient({projectId: 'abc123', perspective: 'published'})).not.toThrow(
```
snorrees, line 165: I would just remove string from the type, and do //@ts-expect-error on these tests.

src/types.ts
stipsan: It was to make it less annoying for people to do:
```ts
const config = {
  projectId: 'abc123',
  dataset: 'production',
  perspective: 'published' as const // <- tried to lose the `as const` req here
}

const client = createClient(config)
```
I'll remove it and re-assess later how to handle this papercut

PR #271, opened 2023-07-31
feat: limit the number of connections open in Node.js

judofyr:
Enable keep-alive, and in addition limit the number of sockets that can be opened. This avoids opening too many connections to the server if someone tries to execute a bunch of requests in parallel. It's recommended to have a concurrency limit at a "higher limit" (i.e. you shouldn't actually execute hundreds of requests in parallel), and this is mainly to minimize the impact for the network and server.

We're currently matching the same defaults as browsers: https://stackoverflow.com/questions/26003756/is-there-a-limit-practical-or-otherwise-to-the-number-of-web-sockets-a-page-op

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@sanity/eventsource": "^5.0.0",
-        "get-it": "^8.3.1",
+        "get-it": "^8.3.0",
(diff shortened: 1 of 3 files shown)


stipsan (approved):
LGTM 👍

rexxars (approved):
Grazie!

PR #273, opened 2023-08-01
feat(projects): `includeMembers` option on `projects.list()`

rexxars:
Recent addition to the API: you can pass a `includeMembers=false` query parameters to `/projects` in order to exclude members list from the response. This reduces the amount of bytes transferred, as well as the response time of the request. Will be used in the CLI tool, amongst others!


stipsan (approved):
Nice addition! 👍

PR #299, opened 2023-08-16
fix: set `useCdn: false` automatically when `perspective: 'previewDrafts'`

stipsan:


src/data/dataMethods.ts
```diff
@@ -323,6 +323,10 @@ export function _requestObservable<R>(
     if (typeof perspective === 'string' && perspective !== 'raw') {
       validateApiPerspective(perspective)
       options.query = {perspective, ...options.query}
+      // If the perspective is set to `previewDrafts` we can't use the CDN, the API will throw
+      if (perspective === 'previewDrafts') {
+        useCdn = false
+      }
```
stipsan: Should we warn when this happens?

E.g. something like
```suggestion
      if (useCdn && perspective === 'previewDrafts') {
        useCdn = false
        console.warn(`perspective: 'previewDrafts' isn't supported with useCdn: true`)
      }
```

rexxars: I think we should. We have a warning printer which ensures we only print once - should probably use that?
https://github.com/sanity-io/client/blob/main/src/warnings.ts#L5

src/warnings.ts
```diff
@@ -12,6 +12,11 @@ export const printCdnWarning = createWarningPrinter([
   `\`useCdn: false\` to use the Live API. Note: You may incur higher costs using the live API.`,
 ])
 
+export const printCdnPrevieDraftsWarning = createWarningPrinter([
+  `The Sanity client is configured with the \`perspective\` set to \`previewDrafts\`, which doesn't support the API-CDN.`,
+  `The Live API will be used instead. Set \`useCdn: false\` in your configuration to hide this warning.`,
```
stipsan, line 17: @rexxars does this look ok?

src/warnings.ts
```diff
@@ -12,6 +12,11 @@ export const printCdnWarning = createWarningPrinter([
   `\`useCdn: false\` to use the Live API. Note: You may incur higher costs using the live API.`,
 ])
 
+export const printCdnPrevieDraftsWarning = createWarningPrinter([
```
rexxars:
```suggestion
export const printCdnPreviewDraftsWarning = createWarningPrinter([
```

src/data/dataMethods.ts
```diff
@@ -323,6 +324,11 @@ export function _requestObservable<R>(
     if (typeof perspective === 'string' && perspective !== 'raw') {
       validateApiPerspective(perspective)
       options.query = {perspective, ...options.query}
+      // If the perspective is set to `previewDrafts` we can't use the CDN, the API will throw
+      if (perspective === 'previewDrafts') {
+        useCdn = false
+        printCdnPrevieDraftsWarning()
```
rexxars:
```suggestion
        printCdnPreviewDraftsWarning()
```

src/warnings.ts
rexxars: Message looks okay to me - suggested a typo fix.

rexxars (approved):

PR #314, opened 2023-09-05
feat(typings): add `StrictUnknownQueryResponseResult` for libraries

stipsan:
Extracted from [`@groqz/json`](https://github.com/stipsan/groqz/blob/main/packages/json/src/index.ts). Every time I work on a GROQ typing POC I find myself reaching for this in the core.

This PR also sets up type checking testing suites using `vitest` and uses it to demonstrate the runtime safety of the new strict response type. We should keep adding more typing test suites so it becomes really easy to avoid introducing unintentionally breaking changes to the TS definitions in the future 🙌

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
       - run: npm ci
       - run: npx ls-engines
       - run: npm run prepublishOnly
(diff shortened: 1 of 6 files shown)


stipsan:
Closing as this ain't it

PR #316, opened 2023-09-05
docs: improve tsdoc for patch method, including filter/params

rexxars:
It was not obvious that you could pass an object of `filter` + `params` to the patch method. This PR improves the situation by using overloads with better documentation.


stipsan:
The build error on ubuntu-latest seems to be a recent hiccup that will probably sort itself after the nodejs core team pushes another nightly, so let's ignore it 🙌

stipsan (approved):
Excellent improvements ✨

bjoerge (approved):

binoy14 (approved):

PR #317, opened 2023-09-05
fix: adjust incorrect listener `visibility` option type

rexxars:
Also improve documentation of listener-related types.

Give that `_listen` is directly exposed on the client as `client.listen`, I also changed the release stage of it to `public`


stipsan (approved):
Excellent ✨

bjoerge (approved):

PR #545, opened 2024-02-14
feat: add support for `returnQuery` option, default to `false`

rexxars:
The `/query` endpoint supports a `returnQuery` option which can be set to `false`. Given that 99% of queries from the client do not use the `filterResponse: false` option, returning this query is unnecessary since it cannot be accessed. Even in the cases where we _do_ use `filterResponse: false` (such as when accessing content source maps), I would argue that it is rare that you actually _need_ the query for anything.

This PR introduces the ability to set the option on a per-query basis, and enables it by default for all queries that do not use the `filterResponse: false` option (for backwards compatibility).

In the next major, I suggest we also disable the query from being returned when using `filterResponse: false`, unless you explicitly also declare `returnQuery: true`.

While this change is fairly trivial, a lot of mocks had to be updated to include the query parameter, so the change appears bigger than it is.

There is a theoretical "breaking" change here _if_ users are relying on mocking specific URLs, but I am not sure it can really thought of as "breaking" per se. Open to opinions here.

Another note: I intentionally did not make it possible to specify `returnQuery: true` if `filterResponse` is not `false`, since there is no way of actually getting at the returned query.

### DIFF
--- a/src/SanityClient.ts
+++ b/src/SanityClient.ts
@@ -28,12 +28,14 @@ import type {
   QueryOptions,
   QueryParams,
   QueryWithoutParams,
+  RawQuerylessQueryResponse,
   RawQueryResponse,
   RawRequestOptions,
   SanityDocument,
   SanityDocumentStub,
   SingleMutationResult,
   UnfilteredResponseQueryOptions,
+  UnfilteredResponseWithoutQuery,
 } from './types'
 import {ObservableUsersClient, UsersClient} from './users/UsersClient'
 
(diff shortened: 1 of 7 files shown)


atombender (approved):
Nice! I'm not familiar with all the types used in the signature, but LGTM.

stipsan (approved):
Great optimisation! 
Maybe we should do a similar optimisation for this case:
```ts
const client = createClient({
  // ...
  resultSourceMap: true || 'withKeyArraySelector'
  // or stega is true, which implicitly sets `resultSourceMap: 'withKeyArraySelector'`
  stega: {
    enabled: true,
    studioUrl: '/studio'
  }
})
// These don't need to fetch Content Source Maps
client.fetch(query, {}, {stega: false, filterResponse: true})
client.fetch(query, {}, {stega: false})
// These still need to fetch CSMs even though they're not externally available as Stega uses them internally
client.fetch(query, {}, {filterResponse: true})
client.fetch(query, {}, {stega: true, filterResponse: true})
// With no stega we should only fetch CSM when filterResponse is false
client.fetch(query, {}, {filterResponse: false})
```

rexxars:
> Maybe we should do a similar optimisation for this case:

I don't have enough overview of the stega/sourcemap feature to opine on this, but you're probably right! Will merge this as-is, and then we can explore further optimizations in separate PRs

PR #560, opened 2024-02-16
feat: optionally encode cross dataset reference specific data

rdunk:
Adds `projectId` and `dataset` to the stega encoded href string via `createEditUrl` if encoding a cross dataset reference. Opt out of this behaviour by setting `omitCrossDatasetReferenceData` to true (is this necessary?).

This is primarily so we can more accurately report documents in use from the visual editing package without the need for augmenting data on the client side.

Also adds a change to return `projectId` and `dataset` from `resolveEditInfo` if present, in case we also want to add these params via `encodeDataAttribute`.

### DIFF
--- a/src/csm/createEditUrl.ts
+++ b/src/csm/createEditUrl.ts
@@ -12,6 +12,8 @@ export function createEditUrl(options: CreateEditUrlOptions): `${StudioBaseUrl}$
     id: _id,
     type,
     path,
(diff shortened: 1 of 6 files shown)


stipsan (dismissed):

rdunk:
Reworked as discussed, description updated.

stipsan (approved):

PR #741, opened 2024-04-17
fix: allow setting middleware on `requester`

stipsan:
This refactor allows setting global middleware, useful for use cases such as logging, or test mocking:

```ts
import {requester} from '@sanity/client'

requester.use({
  onResponse: (response) => {
    console.log('onResponse', response);
    return response;
  },
});

// Later in the app:

const client = createClient({
  projectId: '...',
  apiVersion: '2023-06-21',
  dataset: 'production',
  useCdn: false,
});
await client.fetch(`count(*)`) // logs `onResponse response`
```

Can be tested using:
```bash
npm install @sanity/client@6.15.17-canary.1
```

### DIFF
--- a/src/defineCreateClient.ts
+++ b/src/defineCreateClient.ts
@@ -24,17 +24,10 @@ export default function defineCreateClientExports<
   ClassConstructor: new (httpRequest: HttpRequest, config: ClientConfigType) => SanityClientType,
 ) {
   // Set the http client to use for requests, and its environment specific middleware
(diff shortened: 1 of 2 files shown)


rexxars (approved):

rexxars (changes requested):
Retry tests aren't passing 🤔

judofyr:
The test failures seem to be real: We now no longer propagate the client settings for `retryDelay` and `maxRetries` into the requester. I've opened a separate PR in #742 which I think solves it.

PR #742, opened 2024-04-18
fix: allow setting middleware on `requester`

judofyr:
Alternative solution for #741. Opening PR to get the tests to run.


stipsan (approved):
Thank you 💖

PR #773, opened 2024-05-02
feat: add `stegaClean` method, deprecate `vercelStegaCleanAll`

stipsan:


rexxars (approved):
Looks good to me.
For cleanliness I'd suggest we rename the filename (along with the test) to just `stegaClean`, but no biggie.

PR #783, opened 2024-05-03
fix: prevent listener leak on unsubscribe before eventsource module load

rexxars:
#181 introduced a race condition:

If you unsubscribe from the listener observable before the `@sanity/eventsource` module has been loaded, the `stop()` method would short-circuit, since the `es` variable has not yet been assigned. So once the module did load, the eventsource was set up, but events were practically "thrown away". 

This PR introduces a new `unsubscribed` boolean which we set when unsubscribing (duh), and we then validate it to be `false` _before_ setting up an EventSource.

Also removed one of the calls to `getEventSource()` and assigning the `es`, since it appears to me that calling the `open()` method does the exact same. Let me know if I am overlooking something.

Added a test to ensure the behavior works as intended - run the test on the existing code to see the issue.

### DIFF
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -85,16 +85,13 @@ export function _listen<R extends Record<string, Any> = Record<string, Any>>(
 
   return new Observable((observer) => {
     let es: InstanceType<typeof import('@sanity/eventsource')>
-    getEventSource()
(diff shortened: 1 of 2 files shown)


stipsan (approved):
Thanks, great catch! 🙌

PR #788, opened 2024-05-03
fix(stega): remove try/catch block from `stegaClean`

stipsan:
`@vercel/stega` v1.0.2 (updated in #784) changed the `vercelStegaClean` function to no longer need us to wrap it in a try/catch block.

The test suite removes the `Symbol()` test, this shouldn't cause any issues as it was only there to begin with out of an abundance of caution. The values our users were seeing, that caused problems, were `null` and `undefined`.

While at it I also updated `create-pull-request` actions to set the author as `github-actions`, for the same reason as explained here: https://github.com/sanity-io/pkg-utils/pull/845


christianhg:
Is there anything I can/should do in order to test this change?

stipsan:
@christianhg it's alright, we have a comprehensive testing suite that protects against regressions 👍

christianhg (approved):

PR #797, opened 2024-05-07
feat: add experimental new live events API

stipsan:
This new feature is highly experimental, and is intentionally not documented yet (it might change, and for now we need to test our assumptions).
All relevant surfaces are marked with `@alpha` and a message clearly signaling that it's an experimental feature.
It also requires `apiVersion: 'vX'` at the moment.

### DIFF
--- a/src/data/dataMethods.ts
+++ b/src/data/dataMethods.ts
@@ -257,7 +257,7 @@ export function _dataRequest(
   const useGet = !isMutation && strQuery.length < getQuerySizeLimit
   const stringQuery = useGet ? strQuery : ''
   const returnFirst = options.returnFirst
-  const {timeout, token, tag, headers, returnQuery} = options
+  const {timeout, token, tag, headers, returnQuery, lastLiveEventId} = options
 
   const uri = _getDataUrl(client, endpoint, stringQuery)
 
@@ -274,6 +274,7 @@ export function _dataRequest(
     returnQuery,
     perspective: options.perspective,
     resultSourceMap: options.resultSourceMap,
+    lastLiveEventId: Array.isArray(lastLiveEventId) ? lastLiveEventId[0] : lastLiveEventId,
     canUseCdn: isQuery,
     signal: options.signal,
     fetch: options.fetch,
@@ -373,6 +374,10 @@ export function _requestObservable<R>(
         useCdn = false
         printCdnPreviewDraftsWarning()
       }
+    }
+
+    if (options.lastLiveEventId) {
+      options.query = {...options.query, lastLiveEventId: options.lastLiveEventId}
     }
 
     if (options.returnQuery === false) {
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -171,10 +171,6 @@ export function _listen<R extends Record<string, Any> = Record<string, Any>>(
         .then((eventSource) => {
           if (eventSource) {
             es = eventSource
-            // Handle race condition where the observer is unsubscribed before the EventSource is set up
-            if (unsubscribed) {
-              unsubscribe()
-            }
           }
         })
         .catch((reason) => {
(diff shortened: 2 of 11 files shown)


src/data/dataMethods.ts
```diff
@@ -274,6 +274,7 @@ export function _dataRequest(
     returnQuery,
     perspective: options.perspective,
     resultSourceMap: options.resultSourceMap,
+    lastLiveEventId: Array.isArray(lastLiveEventId) ? lastLiveEventId[0] : lastLiveEventId,
```
stipsan, line 277: The use case for this is explained in the `client.test.ts` suite: https://github.com/sanity-io/client/blob/9da8121a336c70f3771ddb0ef88990722a90c323/test/client.test.ts#L593-L613

src/data/dataMethods.ts
```diff
@@ -375,6 +376,10 @@ export function _requestObservable<R>(
       }
     }
 
+    if (options.lastLiveEventId) {
+      options.query = {...options.query, lastLiveEventId: options.lastLiveEventId}
```
stipsan, line 380: The `lastLiveEventId` changes a lot, and logs are easier to read if each request is the same, with only the live id changing at the tail end of each entry.

src/data/listen.ts
```diff
@@ -171,6 +171,10 @@ export function _listen<R extends Record<string, Any> = Record<string, Any>>(
         .then((eventSource) => {
           if (eventSource) {
             es = eventSource
+            // Handle race condition where the observer is unsubscribed before the EventSource is set up
+            if (unsubscribed) {
+              unsubscribe()
```
stipsan: I'll open a separate PR for this, this race condition was discovered in the new `src/data/live.ts` observable, as it will use the browser provided `EventSource` object instead of waiting for a dynamic import, thus hitting this edge case quite often when testing in `useEffect` in a React app that has StrictMode enabled.

src/data/dataMethods.ts
jonabc: 🤔 It looks like there are a few code paths where user-provided options are passed to this function.  If a user provides `lastLiveEventId` both as part of `options.query` and `options.lastLiveEventId`, then the value in `options.query` will get overwritten.   Or would the values be reconciled in some other way?

Either way, the question of which value is preferred could favor either source property.  Probably just something to document when it's time to document the new feature, so that it's known and expected behavior

src/data/live.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        // If the event has a `data` property, then it`s a MessageEvent emitted by the API and we should forward the error and close the connection
+        if ('data' in evt) {
+          try {
+            const event = parseEvent(evt)
+            observer.error(new Error(event.message, {cause: event}))
+          } catch (err) {
+            observer.error(err)
+          }
```
jonabc: nit: is this catch necessary?  `parseEvent` catches errors as well and returns the error, in which case this catch wouldn't be triggered and `event` on line 43 may not match the expected structure.

jonabc (dismissed):
A few nits, nothing blocking for an initial version. Looks good :shipit: !

atombender (dismissed):
LGTM!

src/data/dataMethods.ts
stipsan: There's no user-facing path that lets you alter `options.query` directly from the outside, it's all handled by internals. The way `lastLiveEventId` is handled is equivalent to how existing options like `options.tags`, `options.returnQuery` and `options.perspective` are handled :)

In other words, the following isn't going to work:
```
client.fetch(`*`, {}, {query: {lastLiveEventId: '123'}})
```
The only entrypoint that works is:
```
client.fetch(`*`, {}, {lastLiveEventId: '123'})
```
As we don't allow setting low level properties like `query`, `body`, `method` and other similar properties that would, if anything, only give you more ways to call the API incorrectly :)

src/data/live.ts
stipsan: Thanks, fixing!

jonabc (approved):

PR #818, opened 2024-05-22
feat: add actions API in client

rdyv:
Add new Actions API implementation: https://www.sanity.io/docs/http-actions

Methods implemented:
1. `create`: A create action creates a new draft document.
2. `replaceDraft`: A replaceDraft action replaces an existing draft document.
3. `edit`: An edit action is used to modify an existing draft document. It applies the given patch to the document referenced by draftId.
4. `delete`: A delete action is used to delete the published version of a document and optionally some (likely all known) draft versions.
5. `discard`: A discard action is used to delete the draft version of a document.
6. `publish`: A publish action is used to publish a draft document. If a published version of the document already exists this is replaced by the current draft document.
7. `unpublish`: An unpublish action is used to retract a published document. If there is no draft version then this is created from the published version.


src/data/action.ts
```diff
@@ -0,0 +1,234 @@
...
+  IdentifiedSanityDocumentStub,
+  PatchOperations,
+  PublishAction,
+  ReplaceDraftAction,
+  UnpublishAction,
+} from '../types'
+import * as validators from '../validators'
+
+/** @internal */
+export class BaseAction {
```
bjoerge: This isn't used anywhere? Lots of great API docs here though, could it be integrated with the types instead? Looks like it needs a bit of revising as it's referring to e.g. `commit()` which doesn't seems to stem from an earlier version of this work.

bjoerge (changes requested):
Great work, Radhe – would love to see some unit tests here as well, if possible.

See also suggestion re: unused file below.

src/types.ts
```diff
@@ -511,6 +511,138 @@ export type Mutation<R extends Record<string, Any> = Record<string, Any>> =
...
+  | EditAction
+  | DeleteAction
+  | DiscardAction
+  | PublishAction
+  | UnpublishAction
+
+/**
+ * Creates a new draft document. The published version of the document must not already exist.
+ * If the draft version of the document already exists the action will fail by default, but
+ * this can be adjusted to either leave the existing document in place or overwrite it.
```
rneatherway:
```suggestion
 * Creates a new draft document. The published version of the document must not already exist.
 * If the draft version of the document already exists the action will fail by default, but
 * this can be adjusted to instead leave the existing document in place.
```

This part of the documentation is out of date as the overwrite behaviour has moved to `replaceDraft`. I've updated the docs.

src/types.ts
```diff
@@ -511,6 +511,138 @@ export type Mutation<R extends Record<string, Any> = Record<string, Any>> =
...
+  ifExists: 'fail' | 'ignore'
+}
+
+/**
+ * Replaces an existing draft document.
+ * At least one of the draft or published versions of the document must exist.
+ */
+export type ReplaceDraftAction = {
+  actionType: 'sanity.action.document.replaceDraft'
+  /** * Draft document ID to replace, if it exists. */
```
rneatherway:
```suggestion

  /** * Draft document ID to replace, if it exists. */
```

Nit: formatting (missing empty line)

src/types.ts
```diff
@@ -884,6 +1016,22 @@ export type TransactionMutationOptions =
   | TransactionAllDocumentsMutationOptions
   | TransactionAllDocumentIdsMutationOptions
 
+/** @internal */
+export type BaseActionOptions = RequestOptions & {
+  transactionId?: string
+  skipCrossDatasetReferenceValidation?: boolean
```
rneatherway, line 1077:
```suggestion
  skipCrossDatasetReferenceValidation?: boolean
  dryRun?: boolean
```

rneatherway (changes requested):

stipsan:
@j33ty quick heads up, I changed the PR title so it uses semantic prefixing. I noticed you use them for the commits 👍 We usually have them in PR titles as well since "squash & merge" is frequently used, and ensures that regardless of merge method the new version that will be published, and the changelog entries, are correct :)

rneatherway (dismissed):
All my comments are addressed. I'll defer to @bjoerge or @stipsan for final review.

src/types.ts
```diff
@@ -511,6 +511,179 @@ export type Mutation<R extends Record<string, Any> = Record<string, Any>> =
...
+
+/**
+ * Creates a new draft document. The published version of the document must not already exist.
+ * If the draft version of the document already exists the action will fail by default, but
+ * this can be adjusted to instead leave the existing document in place.
+ *
+ * @public
+ */
+export type CreateAction = {
+  /**
```
bjoerge, line 534: Any reason not to not include `actionType` on these actions? I can see that it might add a bit of convenience to not have to provide the (rather long) action type when calling `client.action()` and instead do `action({create: {…}})` etc. but I'm not sure if the convenience is worth it. I'm expecting that most people will come from reading the documentation and expect to be able to pass the [examples](https://www.sanity.io/docs/http-actions#be7a440e5a55) there as an argument to `client.action()`

It's also not clear to a reader that you can't add add multiple actions to the same object, e.g.:
`client.action({edit: {...}, unpublish: {...}})`. TypeScript will complain if you do, but not everyone is using TypeScript, and it _will_ work at runtime, but with an unspecified behavior (based on the order of evaluation here https://github.com/sanity-io/client/pull/818/files#diff-b09d77bf2334c027f21d5a51b58875a9d75d7df979ee5ae06fcdd8f8076fa7f7R260). We _could_ validate against it, but I still feel it leads to better DX to be able to use the examples from docs directly here.

src/types.ts
```diff
@@ -511,6 +511,179 @@ export type Mutation<R extends Record<string, Any> = Record<string, Any>> =
   | {delete: MutationSelection}
   | {patch: PatchMutationOperation}
 
+/** @public */
+export type Action =
```
bjoerge, line 515: Re: https://github.com/sanity-io/client/pull/818#discussion_r1611395876
I suggest we make this a union of all the actions, e.g.:
```ts
export type Action =
  | CreateAction
  | ReplaceDraftAction
  | EditAction
  | DeleteAction
  | DiscardAction
  | PublishAction
  | UnpublishAction
```

test/client.test.ts
```diff
@@ -1463,6 +1465,163 @@ describe('client', async () => {
       },
     )
 
+    test.skipIf(isEdge)('action() performs single operation', async () => {
```
bjoerge, line 1474: Thanks for adding these ❤️

src/types.ts
bjoerge: (sorry for not catching this in the first round of reviewing)

rdyv: I did include `actionType` in the initial PR which you reviewed. While writing tests, it felt a bit awkward to include this rather long string which is hard to remember in all the actions, hence rewrote it.

I have reverted the changes for Actions to include `actionType` now.

stipsan (approved):
All good from my perspective 👍 I didn't enable auto merge so that @bjoerge has the opportunity to do the final approval ☺️

bjoerge (approved):

Issue #833, opened 2024-06-03
Explicitly disallow setting both withCredentials and useCdn


amrfarid140:
**Is your feature request related to a problem? Please describe.**
When defining the client configuration you can have the following combination

```
..
"useCdn": true,
"withCredentials": true,
...
```

Since `withCredentials` is set, the client will silently force useCdn to `false`. 

In my project, I've ended up noticing after releasing the website and consuming 80% of my API requests and wish that there was a clear documentation about it and a warning when building the astro project.

**Describe the solution you'd like**
Ideally, the type system should take care of this issue by splitting the type of the configuration into two types. One with cdn (and no withCredentials) and one with withCredentials (and no cdn).

Similar to 

```
type ClientConfiguration = { type: 'mutable', token: string } | { type: 'read-only', useCdn: boolean }
```

Minimum, would be great to have a console warning message highlighting the fact that setting `withCredentials` will force `useCdn` to false.

**Describe alternatives you've considered**
I am currently creating two sanity client instances. One with token and one without `token` + `useCdn` set to true.

amrfarid140:
Here's the related LOC where `useCdn` is forced to `false`. 

https://github.com/sanity-io/client/blob/main/src/config.ts#L134

rexxars:
Thanks for the PR - left some feedback on it 👍

A thing to note here is that the `withCredentials` option is only needed for automatic cookie-based authentication in the browser, which happens primarily when you authenticate through the studio in browsers that allow third-party cookies. In your case, it sounds like you are passing a token - not relying on cookies. In that case, there is no need to pass the  `withCredentials` flag.

PR #834, opened 2024-06-03
fix: add warning about setting both useCdn and withCredentials to true

amrfarid140:
This is a proposed fix to address the minimum solution suggested for #833


stipsan (commented):
Thanks for the PR ❤️  Could you change your commit title, and PR title, to have `fix: add warning about setting both useCdn and withCredentials to true`? That way the release automation works :)

amrfarid140:
Thanks @stipsan ! I've updated the title.

amrfarid140:
Messed up the git history here 😅 . Here's a new PR with the `fix:` attached to both commit and title.

https://github.com/sanity-io/client/pull/849

PR #849, opened 2024-06-10
fix: add warning about setting both useCdn and withCredentials to true

amrfarid140:
This is a proposed fix to address the minimum solution suggested for https://github.com/sanity-io/client/issues/833

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -130,6 +130,11 @@ export const initConfig = (
 
   newConfig.apiVersion = `${newConfig.apiVersion}`.replace(/^v/, '')
   newConfig.isDefaultApi = newConfig.apiHost === defaultConfig.apiHost
+
+  if (newConfig.useCdn === true && newConfig.withCredentials) {
+    warnings.printCdnAndWithCredentialsWarning();
+  }
+
   // If `useCdn` is undefined, we treat it as `true`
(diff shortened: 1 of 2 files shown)


rexxars:
Thanks for the PR! I see the intention behind this, and agree that a warning is warranted. However, the current warning message in this PR is:

> Since you have set `withCredentials` to true, we will default `useCdn` to false.

I don't think this strongly enough conveys what actually happens – "default" sounds to me like "if you haven't set `useCdn` to anything", whereas what happens is actually more along the lines of
> Because you set `withCredentials` to true, we will override your `useCdn` setting to be `false` since (cookie-based) credentials are never set on the CDN

Does that make sense?

I'll make a note that we should update the documentation/TSDoc on the `withCredentials` flag - it generally shouldn't be used, unless you know that you need studio cookie-based authorization headers to be sent along with the request. Perhaps we should also warn when passing both a `token` and `withCredentials: true`, as those indicate two different ways to pass authorization information to the API.

amrfarid140:
Yes the suggested message is more direct!. I've push an update with the new ticket and hopefully a fix for lint.

rexxars (approved):

rexxars:
Thank you @amrfarid140 - this is great ❤️

PR #858, opened 2024-06-19
feat(codegen): Allow query reponse types to be overridden through SanityQueries

romeovs:
This PR is one of two PR's that implements my feature request (https://github.com/sanity-io/sanity/issues/6934) in `sanity-io/sanity`, with the help of some pointers from @sgulseth.

The companion PR that implements the code generation for the newly introduced `SanityQueries` lives at https://github.com/sanity-io/sanity/pull/6997.

It sets up the `@sanity/client` package to have an interface `SanityQueries` that `@sanity/codegen` (or other packages) can hook into assign custom query types automatically based on the queries that are found in the codebase.

This makes it possible to write code that is fully self-typing:

```ts
import {defineQuery} from "groq"
import {createClient} from "@sanity/client"

const client = createClient({ /* ...  */ })

// Define a query as a static string
const posts = defineQuery(`*[_type = "post"] { title, tags }`)

// This will automatically be type correctly as PostsResult
const resp = await client.fetch(query)


// The following will be generated by @sanity/codegen
declare module "@sanity/client" {
  type PostsResult = {
    title?: string
    tags?: string[]
  }

  export interface SanityQueries {
    "*[_type = \"post\"] { title, tags }": PostsResult
  }
}
```

I have tried to set things up so that they are backwards compatible. Users can still manually 
pass the types they want to the query, and the type signatures don't change. I have not touched any existing tests and type tests.

For a way to test the type changes, create a file like the above example and ensure all the types line up with what's expected.

There is also a [TS Playground](https://www.typescriptlang.org/play/?jsx=0#code/PTAEBUAsEsGdTqA7pAhgF2dANt0AjAU1ABNCAzaAO0JIStAAFZUrp0BPYAY22kKqZU5dIQBOodJGIs2nALTQA9jz4D0AYgAcAVi2gA5oXTwAtuKMkAUGV6oxxU0pIBXbMQDkzVuy69+gh6gAN5WoPSiYuSo3MQAyj6cAIou4vzwwQC+VmGghAAeAA5KYpichcQAwmqCAErGLmJUADy54QDiYkoAjnHoYtQGefmiVCTwsP2DADRtoABiqLj4MQDWoAC8oKwcs+EAfJugnT19A1RDBaPjoKuEHErkoAlyHClphPAA-M+Jb6kDT4AbROvSmFwAuqAAFwLJbYFbcVY5cIFYqlCLiaKxX6vaoBTChcLhcjGbiQZq1I47aagbhKKiTUBJYbXCbgoZbSbnAz7AAU3QBHFhSQAlLCAApdUxwQjNfHqeroRotJK02r7fa5bLZKwgCAweAGJSfCJKQxdbo2Qh2BygJyudygABEBktzpCuXILio3HQygYZEoNHeYg4zXpjMw4FZAhu3MG-MF4mFEHFEE9xNADmVTVAybD2qsuv1UEQRhMhgE4gwtAIHCYsl8PGchCMDD5Ui6LgMkEk0kbf1U0FFBFQsDrDP7xALDaI2CUSD1YHQSgr0gkSHYfak8T+oFg5MIplQVnKxAAsv8U-VYG5MFtgqAAPrn2EechKJRBTJAiE5WxsHsRxnDcTxvFeYd1CCIlMSiGI91eUN0kzYkPAAKiBT8lAhDxYSvUMOFve8ixLMAoGIFwJwkbhWFAKhF1AKjiF3fMhQPYRCGwBtqG2SQOAqOgkFQDgrGgUx0UJOkHFrBVBFATJQHIaVQC8JtOCgwIxIkkopKDahCEIhSlJU113RySMmX8dQjm4GTRDk9A+SyUUcmXe1rzDBB4H0mg6FgJRzBQcRiF41jmLEWAgnpMgVgnKxLMwUxPIbLZfMMoU+QwrCv1w1z3LLeAHDvbBMEQc9rES-jBNvI4+VQYT2DpGp0AAOlJdByT5ZLCNFUcWH9WBKFNAihWI0qcn1ABJTAHBPah4Aq7Z4AXC5logATCDiOzoEKTB6rGQdIJitsBGQaQGHbGt-QuUdaIYahSQkVjZwPDk3owaBaNwUSquU1ADHMeStg-L8PAShkmWSgB5XcxCMtKKAMwi+QAA0wgASYJ-sB9RMghVH+s+oaUOy7DcIhqNQCUOHaq2erGswazBHaslIG6jhYY3XqicG4b4FGm9PhI9yAFVYEGadmoJUA+DuAgKBKGR-VwZASlWWBaS3KQlBcWbCEFaAeTpVs6TQC5PkppkfQqunQAZ1AmuZtqOq6jCPD6ji+ZQnYrcwJQADdxAXVAyBIO2HadlrWc6iknxWMRYSoFxTCICRMn5d3PYGuB+dQhOk5TtPiysIA) that summarizes the changes made by this PR and the companion PR.


src/types.ts
```diff
@@ -1236,6 +1236,15 @@ export interface LiveEventMessage {
...
 }
 
+interface QueryReturn {}
+
+export interface SanityQueries extends QueryReturn {}
+
+export type ClientReturn<
+  GroqString extends string,
+  Fallback = Any,
+> = GroqString extends keyof SanityQueries ? SanityQueries[GroqString] : Fallback
```
romeovs, line 1246: @sgulseth this type is slightly different from the type from your tsplayground we discussed on Slack: it is missing the `GeneratedOperations` argument.

It works like this without it and could not find a reason to include it, but I might be overlooking something. Let me know if you want me to include the `GeneratedOperations` argument.

src/types.ts
```diff
@@ -1236,6 +1236,13 @@ export interface LiveEventMessage {
   tags: SyncTag[]
 }
 
+export interface SanityQueries {}
```
romeovs, line 1240: @sgulseth I've omitted the `extends QueryReturn` here. I don't think it's necessary to include it but I might be overlooking something.

sgulseth: Hm. Might be beneficial for the typescript compiler to be sure that it's map when looking up the value, but we could optimize it later 🤔

stipsan (approved):
This is awesome 💖 let's ship 🚀

rdunk (approved):

PR #860, opened 2024-07-02
feat(shard): allow domain sharding

rexxars:
This introduces the concept of domain sharding. It's a workaround for certain clients that have a network setup that disallows use of HTTP2/3. In applications such as the sanity studio, there may be many open network requests at the same time (listeners, queries, mutations etc) which poses a problem when HTTP1 is used, which has a maximum of 6 connections per hostname.

By spreading requests over a number of hostnames (`:projectId.api.r[1-9].sanity.io`), we can work around that limitation. Note that this prevents any auth cookie from being attached to the request - if you need auth, you'll have to use the authorization header (through the `token` parameter).

We use a "global" sharder singleton instance to avoid multiple different clients having separate buckets and ending up with one bucket being "filled up" because of an unshared state.

Had to take a few shortcuts in the tests here - ideally this would only be applied in browser environments, but couldn't find a great way to do so. We can refactor later if we find a workaround, but it shouldn't be a blocker.

Note that this is implemented _primarily_ as a get-it middleware, but because the `listen` code-path does not use get-it under the hood, it has a separate, manual tracking implementation.

Note: the shard domains may not  be up and running yet.

### DIFF
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -1,5 +1,6 @@
 import {Observable} from 'rxjs'
 
+import {domainSharder as sharder} from '../http/domainSharding'
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
 import type {Any, ListenEvent, ListenOptions, ListenParams, MutationEvent} from '../types'
 import defaults from '../util/defaults'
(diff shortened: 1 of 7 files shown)


binoy14 (approved):
Changes makes sense, was not able to test manually but unit tests look good enough. Thanks!

bruno-oliveira-fanduel (approved):

PR #872, opened 2024-07-19
fix: add support for includeMutations listen parameter

bjoerge:
Our `/listen` endpoint supports passing `includeMutations=false` for opting out of receiving mutations, which reduces bandwidth in cases where mutations are not relevant. This adds support for passing it as an option when setting up a listener, example:

```js
client.listen('*', {}, {includeMutations: false})
```


rdyv (approved):
LGTM

rexxars (approved):

PR #879, opened 2024-08-13
fix: deprecate studioHost, externalStudioHost in typings

rexxars:
Should be pretty self explanatory 

Fixes SDX-1494


binoy14 (approved):
Thanks for the article as well

stipsan (approved):

PR #881, opened 2024-08-14
fix: support `signal` on `getDocument(s)` to cancel requests

stipsan:
Allows cancelling `getDocument` and `getDocuments` with an `AbortController` when using the promise API:

```ts
import {useEffect} from 'react'

const client = createClient({ /* ... */ })

const [document, setDocument] = useState(null)
useEffect(() => {
  const controller = new AbortController()
  client.getDocument('abc123', {signal: controller.signal}).then(setDocument)
  return () => controller.abort()
}, [])
```

The `ObservableSanityClient` doesn't need this, as it automatically cancels inflight requests when the observer no longer has any subscribers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@sanity/client",
-  "version": "6.21.2",
+  "version": "6.21.3-canary.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


christianhg (approved):

rdunk (approved):
LGTM!

PR #894, opened 2024-10-03
fix: add missing `listenerName` property on welcome event

rexxars:
The `listenerName` property is handy to debug certain listener related issues, but was not part of the typings.


binoy14 (approved):
Makes sense to me, thanks!

ricokahler (approved):

PR #901, opened 2024-10-25
docs: add reference docs for client.live.events()

kmelve:
Adds a section on the new and experimental `client.live.events()` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -887,7 +887,7 @@ client.getDocuments(ids).then((docs) => {
 
 > [!NOTE]
 >
-> Live Content is experimental and require your client config to be set up with `apiVersion: 'vX'`.
+> Live Content is experimental and requires your client config to be set up with `apiVersion: 'vX'`.
 
 ```ts
 // Subscribe to live updates
--- a/README.md
+++ b/README.md
@@ -883,6 +883,47 @@ client.getDocuments(ids).then((docs) => {
   // [{_id: 'bike123', ...}, null, {_id: 'bike345', ...}]
 })
 ```
+### Listening to live content updates
+
+> [!NOTE]
+>
+> Live Content is experimental and require your client config to be set up with `apiVersion: 'vX'`.
+
+```ts
+// Subscribe to live updates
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -883,6 +883,47 @@ client.getDocuments(ids).then((docs) => {
   // [{_id: 'bike123', ...}, null, {_id: 'bike345', ...}]
 })
 ```
+### Listening to live content updates
+
+> [!NOTE]
+>
+> Live Content is experimental and require your client config to be set up with `apiVersion: 'vX'`.
```
markmichon:
```suggestion
> Live Content is experimental and requires your client config to be set up with `apiVersion: 'vX'`.
```

README.md
```diff
@@ -883,6 +883,47 @@ client.getDocuments(ids).then((docs) => {
...
+  (event) => { 
+    // Check if incoming tags match saved sync tags 
+    if (event.type === "message" && event.tags.some((tag) => syncTags.includes(tag))) { 
+      // Refetch with ID to get latest data
+      render(event.id)
+    }
+    if (event.type === "restart") {
+      // A restart event is sent when the `lastLiveEventId` we've been given earlier is no longer usable
+      render()
+    }
```
markmichon, line 904: I'm not sure how much sense this makes out of context.  Maybe it needs to be more clear this is a rough example of subscribing (more vague), or it needs to include the surrounding context like the render function, receiving syncTags from the response (more specific)

kmelve: Yeah, I wasn't sure how much we should go into it here since the convention is to make the code examples fairly light-weight. Maybe we should just link to the docs for implementation details.

stipsan: It's still useful and should be included, the most common case where a `restart` will happen is if you lose your internet connection and are offline long enough, and the live events that happen while you are gone are many enough, that the API can't replay everything that happened. So instead of replying the events, you'll get the `restart` and should refetch as it's highly likely the content you're looking at is stale.

PR #902, opened 2024-10-30
[WIP] feat(client.fetch): add support for bundlePerspective

pedrobonamin:
Adds support for `bundlePerspective` in `client.fetch`

Bundle perspective allows users to define a list of bundle ids which they want to query data from. 
Content Lake will infer how the documents should be returned when using this new perspective.


bjoerge:
We can close this now, right @pedrobonamin ?

PR #910, opened 2024-11-05
fix(live): detect CORS errors

stipsan:
Detects CORS errors and throws a custom error:
```bash
CorsOriginError: The current origin is not allowed to connect to the Live Content API. 
Add it here: https://sanity.io/manage/project/hiomol4a/api?cors=add&origin=https%3A%2F%2Fexample.com
```

Go to https://example.com and run this snippet in your console to see the error for yourself:
```js
const {createClient} = await import('https://esm.sh/@sanity/client@canary')

const client = createClient({
    projectId: 'hiomol4a',
    dataset: 'fashion-show',
    useCdn: true,
    apiVersion: 'X'
})

client.live.events().subscribe({next: (event) => console.log(event), error: err => console.error(err)})
```

Userland can use the `CorsOriginError` export to handle CORS differently than other errors:

```ts
import {CorsOriginError, createClient} from '@sanity/client'

const client = createClient({
  // ...
})

client.live.events().subscribe({
  next: (event) => {
    // ...
  },
  error: (err) => {
    if(err instanceof CorsOriginError) {
      toast.warn(err.message)
    } else {
      toast.error(err)
    }
  }
})
```


test/live.test.ts
```diff
@@ -24,6 +27,29 @@ const testSse = async (onRequest: OnRequest, options: ClientConfig = {}) => {
 describe.skipIf(typeof EdgeRuntime === 'string' || typeof document !== 'undefined')(
   '.live.events()',
   () => {
+    const server = setupServer(
```
stipsan, line 30: As explained further down, `nock` isn't able to intercept `global.fetch`, so we use `msw` for this.
We might be able to use `msw` instead of `nock`, but it feels out of scope for this PR.

test/live.test.ts
```diff
@@ -214,6 +240,81 @@ describe.skipIf(typeof EdgeRuntime === 'string' || typeof document !== 'undefine
...
+      expect.assertions(3)
+      const restoreFetch = global.fetch
+
+      global.fetch = async (info, init) => {
+        const response = await restoreFetch(info, init)
+        if (!response.headers.has('access-control-allow-origin')) {
+          throw new Error('CORS preflight request failed')
+        }
+        return response
+      }
```
stipsan, line 253: `global.fetch` is based on `undici` in node, which doesn't actually implement CORS, so we have to resort to some trickery to mimic the behaviour we want.

PR #925, opened 2024-11-20
feat(request): add flag to disable logging api warnings

dcilke:
When using the client library and the `vX` api, it can be undesirable/noisy to log
```
This is a future API version, which may change without warning.
```
for every request.

### DIFF
--- a/src/defineCreateClient.ts
+++ b/src/defineCreateClient.ts
@@ -24,12 +24,13 @@ export default function defineCreateClientExports<
   ClassConstructor: new (httpRequest: HttpRequest, config: ClientConfigType) => SanityClientType,
 ) {
   // Set the http client to use for requests, and its environment specific middleware
-  const defaultRequester = defineHttpRequest(envMiddleware)
+  const defaultRequester = defineHttpRequest(envMiddleware, undefined)
 
-  const createClient = (config: ClientConfigType) =>
-    new ClassConstructor(
+  const createClient = (config: ClientConfigType) => {
+    const configRequester = defineHttpRequest(envMiddleware, config)
+    return new ClassConstructor(
       (options, requester) =>
-        (requester || defaultRequester)({
+        (requester || configRequester)({
           maxRedirects: 0,
           maxRetries: config.maxRetries,
           retryDelay: config.retryDelay,
@@ -37,6 +38,7 @@ export default function defineCreateClientExports<
         } as Any),
       config,
     )
+  }
 
   return {requester: defaultRequester, createClient}
 }
--- a/src/defineCreateClient.ts
+++ b/src/defineCreateClient.ts
@@ -24,13 +24,13 @@ export default function defineCreateClientExports<
   ClassConstructor: new (httpRequest: HttpRequest, config: ClientConfigType) => SanityClientType,
 ) {
   // Set the http client to use for requests, and its environment specific middleware
-  const defaultRequester = defineHttpRequest(envMiddleware, undefined)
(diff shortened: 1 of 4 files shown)


src/http/request.ts
```diff
@@ -17,21 +17,25 @@ const httpError = {
...
 }
 
 /** @internal */
-export function defineHttpRequest(envMiddleware: Middlewares): Requester {
+export function defineHttpRequest(envMiddleware: Middlewares, config?: ClientConfig): Requester {
   return getIt([
     retry({shouldRetry}),
     ...envMiddleware,
-    printWarnings,
+    printWarnings(config?.ignoreSanityAPIWarnings),
```
rdyv: Another alternative here is to skip `printWarnings` middleware if `config.ignoreSanityAPIWarnings` flag is set. That makes `printWarnings` middleware simpler.

judofyr:
How about making the setting: `ignoredAPIWarnings: string[]` and yo need to do `ignoredAPIWarnings: ["This is a future API version, which may change without warning."]`? That way you end up not ignoring other potential API warnings that could appear in the future?

dcilke:
> How about making the setting: `ignoredAPIWarnings: string[]` and yo need to do `ignoredAPIWarnings: ["This is a future API version, which may change without warning."]`? That way you end up not ignoring other potential API warnings that could appear in the future?

I guess that speaks to the intent of the flag. I'm thinking of this more as a log level filter. Since that isn't implemented, I opted for the existing pattern of an `ignore...` flag to mimic that behavior.

bjoerge:
> ignoredAPIWarnings: string[]

Agree with this - would be good to be able to opt out of particular warning(s), and still be able to get others. Guess that requires all warnings to be static strings (I suppose we don't have identifiers for these?), or allow providing a regex to match the warnings you want to ignore.

Might also be worth to consider skipping warnings that has already been logged?

src/defineCreateClient.ts
```diff
@@ -24,13 +24,13 @@ export default function defineCreateClientExports<
   ClassConstructor: new (httpRequest: HttpRequest, config: ClientConfigType) => SanityClientType,
 ) {
   // Set the http client to use for requests, and its environment specific middleware
-  const defaultRequester = defineHttpRequest(envMiddleware, undefined)
+  const defaultRequester = defineHttpRequest(envMiddleware)
 
   const createClient = (config: ClientConfigType) => {
-    const configRequester = defineHttpRequest(envMiddleware, config)
+    const clientRequester = defineHttpRequest(envMiddleware)
```
bjoerge, line 30: Is it intentional to not pass config here anymore?

dcilke: Yes, based on the feedback above, I'm opting for a "warn once" pattern. https://github.com/sanity-io/client/pull/925/files#diff-e54beffa7132ee4fc0c37a1b8d3fa97bce3a4aea12db7ec23fdd28c112b3c05bR21-R33

I'm still thinking through how the warning ought to be. To me, it makes sense to limit version warnings to once as they are set on the client. However, we also use the warning header for duplicate parameters which should probably be always printed.

bjoerge: But isn't this a breaking change? If you call `createClient(myConfig)`, won't it now entirely ignore `myConfig`?

bjoerge: Ah, I guess it's still being applied to the ClassConstructor call further down

bjoerge (approved):

PR #934, opened 2024-12-03
feat(query): add support for release perspectives

bjoerge:
This opens up for passing multiple perspectives to `client.fetch()` et al., enabling support for releases.

So far, only additional perspectives we support must be a ReleaseId, which by convention must start with the character `r`.

See unit tests for more details.

Multiple perspectives will be applied in order of significance.

e.g. `perspective=rABC,drafts` will return the following under the given scenarios:
- Dataset includes documents `[{_id: "foo", …}, {_id: "drafts.foo", …}, {_id: "versions.rABC.foo", …}]`
  - returns `[{_id: "versions.rABC.foo", …}]`
- Dataset includes documents `[{_id: "foo", …}, {_id: "drafts.foo", …}]`
  - returns `[{_id: "drafts.foo", …}]`
- Dataset includes documents `[{_id: "foo", …}]`
  - returns `[{_id: "foo", …}]`(since published documents are included by default)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -28,7 +28,28 @@ function validateApiVersion(apiVersion: string) {
   }
 }
 
-export const validateApiPerspective = function validateApiPerspective(perspective: string) {
+export const validateApiPerspective = function validateApiPerspective(perspective: unknown) {
+  if (Array.isArray(perspective)) {
+    for (const perspectiveValue of perspective) {
+      if (perspectiveValue === 'published') {
+        continue
+      }
+      if (perspectiveValue === 'drafts') {
+        continue
+      }
+      if (
+        typeof perspectiveValue === 'string' &&
+        perspectiveValue.startsWith('r') &&
+        perspectiveValue !== 'raw'
+      ) {
+        continue
+      }
+      throw new TypeError(
+        'Invalid API perspective value, expected `published`, `drafts` or a valid release identifier string',
+      )
+    }
+    return
+  }
   switch (perspective as ClientPerspective) {
     case 'previewDrafts':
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -74,7 +95,7 @@ export const initConfig = (
     throw new Error('Configuration must contain `projectId`')
   }
 
-  if (typeof newConfig.perspective === 'string') {
+  if (typeof newConfig.perspective !== 'undefined') {
```
bjoerge, line 99: figured it's better if we always validate when it's provided. As it was, we would silently ignore `perspective={}` for example, which is a mistake I would appreciate being told about :)

pedrobonamin (dismissed):
Thanks for doing this, looks good to me

src/config.ts
```diff
@@ -28,7 +28,28 @@ function validateApiVersion(apiVersion: string) {
...
+        continue
+      }
+      throw new TypeError(
+        'Invalid API perspective value, expected `published`, `drafts` or a valid release identifier string',
+      )
+    }
+    return
+  }
   switch (perspective as ClientPerspective) {
     case 'previewDrafts':
```
rneatherway, line 54: I'd suggest also allowing `drafts` as a string parameter.

bjoerge: Ah, yes, good point. Added in https://github.com/sanity-io/client/commit/9f105c259f6ff2c41133fa9255f6b8cfbc77e57d

rneatherway (approved):

stipsan (approved):
LFGO

PR #949, opened 2025-01-03
refactor(test): cleanup observable tests

bjoerge:
This simplifies the tests for client.listen and client.live, making them less verbose and (hopefully) a bit easier to follow.

Also added a test to assert that the client sends `last-event-id` header in case of a disconnect.

Also fixed a couple of warnings that was issued during test runs due to lack of awaiting (or returning) promises at the end of a few tests:
```
Promise returned by `expect(actual).rejects.toBeDefined()` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in Vitest 3. Please remember to await the assertion.
    at client/test/client.test.ts:429:46
```
++ a couple of other minor improvements

### DIFF
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -20,7 +20,7 @@ import {
   Transaction,
   UnpublishAction,
 } from '@sanity/client'
-import {of as observableOf} from 'rxjs'
+import {firstValueFrom, lastValueFrom, of as observableOf, toArray} from 'rxjs'
 import {filter} from 'rxjs/operators'
 import {describe, expect, test, vi} from 'vitest'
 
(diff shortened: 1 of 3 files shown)


test/client.test.ts
```diff
@@ -2600,7 +2576,7 @@ describe('client', async () => {
         'data: {"listenerName":"LGFXwOqrf1GHawAjZRnhd6"}',
         '',
         'event: mutation',
-        `data: ${JSON.stringify({document: doc})}`,
+        `data: ${JSON.stringify({result: doc})}`,
```
bjoerge, line 2579: this aligns with mutation events from /listen endpoint returning `result`, not `document`

test/listen.test.ts
```diff
@@ -140,6 +116,37 @@ describe.skipIf(typeof EdgeRuntime === 'string' || typeof document !== 'undefine
       }
     })
 
+    test('sends last-event-id header when reconnecting', async () => {
```
bjoerge, line 119: buried lede, perhaps, but this change is in a separate commit, so we might want merge with rebase to keep this as a separate item in the commit history.

stipsan (approved):
Much easier to follow indeed 💖

PR #950, opened 2025-01-03
refactor: unify sse endpoints handling

bjoerge:
This needs a bit of a close look, but tldr; it refactors our current EventSource/SSE handling and unifies it under a common `connectEventsource()` function, now shared between the Live Content API and the Listener API. This function handles Sanity API SSE endpoint specific behaviors like dealing with `channelError`, `disconnect` event etc, but should be flexible enough to adapt to potential new future SSE endpoints.

The `connectEventsource()` function takes in a `initEventSource` function that allows us to abstract out the actual eventsource instantiation, which tends to vary depending on request options (headers) and environments.

Here's an example of how live query api uses it to determine which eventsource implementation to use:
https://github.com/sanity-io/client/blob/b07b8d626a039f8c30ea58a2f2ae28d8b9a0e18c/src/data/listen.ts#L96

This will lazy load the polyfill if either 1) no global EventSource is defined or 2) if headers are passed. Otherwise it will use the native EventSource class provided by the runtime.

I've also formalized the error classes and documented what errors can be thrown on the observable and under what circumstances they will be thrown. This should enable downstream consumers to better reason about and recover from the different types of errors.

I feel fairly confident that this patch doesn't introduce any breaking changes (beyond the bugfix mention below). Please tell me if I've overlooked something else.

This also exports the standalone `connectEventsource()` and the different error classes/helper types changes, which enables internal usage (tagged as `@internal` for now).

### Backwards incompatible change with regards to `disconnect` events

Currently, if the server sends a `disconnect` event, the listener stream will _complete_. This is from `main`: https://github.com/sanity-io/client/blob/5f0a991a2f72632454be7489146c973399008c5d/src/data/listen.ts#L133

This seems like a bug with a potential for unfortunate effects. Consider the following:

1. A listener is set up to receive patches for a certain document. The returned observable emits mutations and the application use these to keep a local version of the document in sync.
2. The backend sends a `disconnect` event, which causes the observable to complete, and never reconnect.
No error is thrown and the application is not likely to pick up on the error and continue as if nothing happened, leaving the document out of sync.

I'd argue that this is a bug, and that the observable here should be kept "open" forever until the subscriber unsubscribes, and that a `disconnect` event should be treated as an error condition. Now, it looks like the only cases where a `disconnect` is actually sent are cases where the dataset has been deleted or similar exceptional cases, but I'd argue that throwing an error on the observable in these cases still would be preferable, and doing so is not a breaking change, but should be considered a bugfix with the reasoning that _if it happens_, we actually want it to raise an error instead of silently fail.

### DIFF
--- a/src/data/eventsource.ts
+++ b/src/data/eventsource.ts
@@ -1,0 +1,255 @@
+import {defer, isObservable, mergeMap, Observable, of} from 'rxjs'
+
+import {type Any} from '../types'
+
+/**
+ * @public
+ * Thrown if the EventSource connection could not be established.
+ * Note that ConnectionFailedErrors are rare, and disconnects will normally be handled by the EventSource instance itself and emitted as `reconnect` events.
+ */
+export class ConnectionFailedError extends Error {
+  readonly name = 'ConnectionFailedError'
+}
+
+/**
+ * The listener has been told to explicitly disconnect.
+ *  This is a rare situation, but may occur if the API knows reconnect attempts will fail,
+ *  eg in the case of a deleted dataset, a blocked project or similar events.
+ * @public
+ */
+export class DisconnectError extends Error {
+  readonly name = 'DisconnectError'
+  readonly reason?: string
+  constructor(message: string, reason?: string, options: ErrorOptions = {}) {
+    super(message, options)
+    this.reason = reason
+  }
+}
+
+/**
+ * @public
+ * The server sent a `channelError` message. Usually indicative of a bad or malformed request
+ */
+export class ChannelError extends Error {
+  readonly name = 'ChannelError'
+  readonly data?: unknown
+  constructor(message: string, data: unknown) {
+    super(message)
+    this.data = data
+  }
+}
+
+/**
+ * @public
+ * The server sent an `error`-event to tell the client that an unexpected error has happened.
+ */
+export class MessageError extends Error {
+  readonly name = 'MessageError'
+  readonly data?: unknown
+  constructor(message: string, data: unknown, options: ErrorOptions = {}) {
+    super(message, options)
+    this.data = data
+  }
+}
+
+/**
+ * @public
+ * An error occurred while parsing the message sent by the server as JSON. Should normally not happen.
+ */
+export class MessageParseError extends Error {
+  readonly name = 'MessageParseError'
+}
+
+/**
+ * @public
+ */
+export interface ServerSentEvent<Name extends string> {
+  type: Name
+  id?: string
+  data?: unknown
+}
+
+// Always listen for these events, no matter what
+const REQUIRED_EVENTS = ['channelError', 'disconnect']
+
+/**
+ * @internal
+ */
+export type EventSourceEvent<Name extends string> = ServerSentEvent<Name>
+
+/**
+ * @internal
+ */
+export type EventSourceInstance = InstanceType<typeof globalThis.EventSource>
+
+/**
+ * Sanity API specific EventSource handler shared between the listen and live APIs
+ *
+ * Since the `EventSource` API is not provided by all environments, this function enables custom initialization of the EventSource instance
+ * for runtimes that requires polyfilling or custom setup logic (e.g. custom HTTP headers)
+ * via the passed `initEventSource` function which must return an EventSource instance.
+ *
+ * Possible errors to be thrown on the returned observable are:
+ * - {@link MessageError}
+ * - {@link MessageParseError}
+ * - {@link ChannelError}
+ * - {@link DisconnectError}
+ * - {@link ConnectionFailedError}
+ *
+ * @param initEventSource - A function that returns an EventSource instance or an Observable that resolves to an EventSource instance
+ * @param events - an array of named events from the API to listen for.
+ *
+ * @internal
+ */
+export function connectEventSource<EventName extends string>(
+  initEventSource: () => EventSourceInstance | Observable<EventSourceInstance>,
+  events: EventName[],
+) {
+  return defer(() => {
+    const es = initEventSource()
+    return isObservable(es) ? es : of(es)
+  }).pipe(mergeMap((es) => connectWithESInstance(es, events))) as Observable<
+    ServerSentEvent<EventName>
+  >
+}
+
+/**
+ * Provides an observable from the passed EventSource instance, subscribing to the passed list of names of events types to listen for
+ * Handles connection logic, adding/removing event listeners, payload parsing, error propagation, etc.
+ *
+ * @param es - The EventSource instance
+ * @param events - List of event names to listen for
+ */
+function connectWithESInstance<EventTypeName extends string>(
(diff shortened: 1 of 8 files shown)


src/data/eventsource.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * @param es - The EventSource instance
+ * @param events - List of event names to listen for
+ */
+function connectWithESInstance<EventTypeName extends string>(
+  es: EventSourceInstance,
+  events: EventTypeName[],
+) {
+  return new Observable<EventSourceEvent<EventTypeName>>((observer) => {
+    const emitOpen = (events as string[]).includes('open')
```
bjoerge, line 128: this now makes it possible to subscribe to `open` events as well (but opt-in)

src/data/eventsourcePolyfill.ts
```diff
@@ -0,0 +1,7 @@
+import {defer, shareReplay} from 'rxjs'
+import {map} from 'rxjs/operators'
+
+export const eventSourcePolyfill = defer(() => import('@sanity/eventsource')).pipe(
```
bjoerge, line 4: this will lazy load the polyfill, share it between consumers and cache it forever

test/listen.test.ts
```diff
@@ -74,16 +74,12 @@ describe.skipIf(typeof EdgeRuntime === 'string' || typeof document !== 'undefine
         ({request, channel}) => {
           expect(request.headers.authorization, 'should send token').toBe('Bearer foobar')
 
-          channel!.send({event: 'disconnect'})
-          process.nextTick(() => {
-            channel!.close()
```
bjoerge, line 79: closing here will trigger an error/reconnecting state, so not needed since we only take the welcome event from the observable

test/listen.test.ts
```diff
@@ -168,8 +164,6 @@ describe.skipIf(typeof EdgeRuntime === 'string' || typeof document !== 'undefine
 
       const {server, client} = await testSse(({channel}) => {
         channel!.send({event: 'channelError', data: {error: {description: 'Expected error'}}})
-        channel!.close()
```
bjoerge, line 171: also not needed since an error thrown on the observable will call the observable cleanup, which will do the closing for us

test/listen.test.ts
```diff
@@ -168,8 +164,6 @@ describe.skipIf(typeof EdgeRuntime === 'string' || typeof document !== 'undefine
...
-        process.nextTick(() => channel!.close())
```
bjoerge, line 172: also not needed since an error thrown on the observable will call the observable cleanup, which will do the closing for us

src/types.ts
```diff
@@ -854,6 +854,15 @@ export type ReconnectEvent = {
   type: 'reconnect'
 }
 
+/**
+ * The listener connection has been established
+ * note: it's usually a better option to use the 'welcome' event
+ * @public
+ */
+export type OpenEvent = {
```
sgulseth, line 862: Should this be part of ListenEvent on L878?

src/data/eventsource.ts
```diff
@@ -0,0 +1,255 @@
...
+ * - {@link ChannelError}
+ * - {@link DisconnectError}
+ * - {@link ConnectionFailedError}
+ *
+ * @param initEventSource - A function that returns an EventSource instance or an Observable that resolves to an EventSource instance
+ * @param events - an array of named events from the API to listen for.
+ *
+ * @internal
+ */
+export function connectEventSource<EventName extends string>(
```
sgulseth, line 104: nit:
```suggestion
const EventTypeName = ListenEvent['type']
export function connectEventSource<EventName extends EventTypeName[]>(
```

should make this type safe? 🤔  and same on L123?

sgulseth (dismissed):
nice, code wise lgtm! some nits around typings 🤷  Would be nice to get either @rexxars or @stipsan feedback as well, as i'm not that familiar with the client

src/data/eventsourcePolyfill.ts
stipsan: Really cool! And so much more elegant than the old `new Promise` song and dance 🤯

stipsan (dismissed):
Absolute brilliant 💖 I always learn so much about `rxjs` when reviewing your PRs, and your improvements are so elegant :chefs-kiss:

src/data/live.ts
stipsan, line 1: What do you think about using a shareReplay cache trick like the polyfill here?
The cache key can be: `JSON.stringify({url: url.href, esOptions})` and it would allow interesting improvements to Sanity Live. Unlike `client.listen()`, the resulting URL doesn't have filters or anything else that lead to many unique URLs. If `includeDrafts` or a `token` isn't set, then the same URL is used for an entire dataset and apiVersion.
For example in Next.js apps it would let you render client components that might call `client.live.events()` multiple times on the same page, and still end up with a single Event Source connection. Free from the burden of implementing this dedupe logic in userland ❤

src/types.ts
bjoerge: Good catch! Will add it

src/data/eventsource.ts
bjoerge: It is type safe… kind of. Here we can only really know for _sure_ that the event matches any of the event names given, which is why this returns `{type: Name, id:? string, data?: unknown}`. I left it up to the endpoints themselves (`listen()` and `live()` – for now) to take care of typing it correctly, since they have more specific knowledge of the payloads. I also made a choice not to splat `data` into the events at this level, because in theory it could contain `type` and `id` which would lead to very surprising behavior.

src/data/live.ts
bjoerge: Not opposed to it, but it's also something we could provide as an operator or small wrapper util. It feels like request deduping is something that belongs more on the application/SDK side though

stipsan: it's more like instance deduping. the client is already scoped to the project and dataset, so the SSE resource feels like something that should be scoped as well.
Both the Listen API and the Live API have limits to how concurrent things are (starts at 1k): https://arc.net/l/quote/edfhjttg

Listen has been difficult to use, but even then we've run into issues where you reach too many editors working in a studio.
With the Live API being so easy, this will grow. Both for production use cases, but for plugins it's become my preferred way of building things where I want to write to a document and subscribe to changes in a very lightweight way that lives outside of a form builder state.
If you install 10 plugins and they all use this technique, but we don't dedupe, you'll have easily 10 connections per user.
And it feels odd to add a wrapper like this in `useClient` in `sanity`, and then in `next-sanity`. `hydrogen-sanity` will also need it. And probably `@sanity/astro`. And what about `@nuxtjs/sanity`. 🤔

sgulseth (approved):

src/data/live.ts
bjoerge: Fair and convincing points. For debuggability it might be worth considering having an option/env var to turn the caching behavior off.

PR #951, opened 2025-01-03
refactor: enable verbatim module syntax

bjoerge:
Verbatim module syntax enforces explicit type imports/exports which ensures predictable compilation. [Details](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax)

### DIFF
--- a/src/csm/types.ts
+++ b/src/csm/types.ts
@@ -1,6 +1,6 @@
 import type {ContentSourceMap, ContentSourceMapDocuments, SanityDocument} from '@sanity/client'
 
-import {Path} from './studioPath'
+import {type Path} from './studioPath'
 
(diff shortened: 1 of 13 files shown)


stipsan (approved):
Agreed, it's why it's in the `"extends": "@sanity/pkg-utils/tsconfig/strict.json"` [preset](https://github.com/sanity-io/pkg-utils/blob/287311988f5f3ed11c1b6c7dca7de2975dc2650b/tsconfig/strict.json#L8).
Since this is such an important package to us, it might be worth using `@sanity/pkg-utils/tsconfig/strictest.json` 🤔

bjoerge:
> Since this is such an important package to us, it might be worth using @sanity/pkg-utils/tsconfig/strictest.json

+1 to this

PR #964, opened 2025-01-09
fix: allow mutation selection without passing a patch builder

nkgentile:
Fixes a bug when using the mutation selector syntax in a transaction, and adjusts the types and overloaded `patch` method.

Previously, this would work as expected (with type errors):
```js
client.transaction().patch(
  {
    query: '*[_id in $ids]',
    params: {ids: ['abc123', 'foo.456']},
  },
  (p) => p.inc({count: 1}),
)
```

But this wouldn't:
```js
client.transaction().patch(
  {
    query: '*[_id in $ids]',
    params: {ids: ['abc123', 'foo.456']},
  },
  {inc: {count: 1}},
)
```

### DIFF
--- a/src/data/transaction.ts
+++ b/src/data/transaction.ts
@@ -7,6 +7,7 @@ import type {
   IdentifiedSanityDocumentStub,
   MultipleMutationResult,
   Mutation,
+  MutationSelection,
   PatchOperations,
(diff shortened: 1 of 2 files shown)


cngonzalez (approved):
This seems reasonable to me!

PR #976, opened 2025-01-21
feat(csm): add document id utils

stipsan:
The code is originally from https://github.com/sanity-io/sanity/blob/9cac7ffdfb1922119c7848bc555e18e4d3df5981/packages/sanity/src/core/util/draftUtils.ts#L5-L23, the idea is for `@sanity/client/csm` to become the source of truth for logic like `getDraftId`.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sanity/client",
-  "version": "6.25.0",
+  "version": "6.25.1-canary.0",
```
pedrobonamin, line 3: Is this expected?

pedrobonamin (approved):
Looks good to me, thanks @stipsan !

package.json
stipsan: yep, it's switched back on release 👍

PR #990, opened 2025-01-27
fix(live): dedupe event source instances

stipsan:


src/data/live.ts
```diff
@@ -118,7 +126,14 @@ export class LiveClient {
         throw new CorsOriginError({projectId: projectId!})
       }),
     )
-    return concat(checkCors, events)
+    const observable = concat(checkCors, events).pipe(
+      shareReplayLatest({
+        predicate: (event) => event.type === 'welcome',
+      }),
+      finalize(() => eventsCache.delete(key)),
```
bjoerge: This should be moved before shareReplay since we want it to happen when finalizing the _shared_ observable

bjoerge (changes requested):

bjoerge (approved):
Thanks for adding tests too! <3

PR #996, opened 2025-01-28
chore(internal): adds support for resource configuration

snorrees:
## [THIS IS](https://github.com/sanity-io/client/pull/1040)

Temp branch to explore resource targets in the api.

Current tagged release: 6.28.0-resources.0

See  added unit-tests tests for what I've tested.

## Caveats
Current impl takes `experimental_resource` or `projectId` `resource.r<esourceType>.<resourceId>` (which then sets `experimental_resource` config)

The client then is forced into `useProjectHostname: false`

 Then _all_ requests get /<resourceType>/<resourceId> prepended (urls are changed)

This is probably not what we want for `client.request` that currently use `useProjectHostname`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ umd
 # MacOS
 .DS_Store
 tsconfig.vitest-temp.json
+
+# JetBrains stuff
+.idea
+*.iml
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sanity/client",
-  "version": "6.28.0",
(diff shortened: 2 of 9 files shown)


sgulseth (commented):
Some general feedback on the configuration;
We need a way to configure resources, not everything should be able to be a resource. We also need to have list of features for each resource, as not every resource will be available to every resource. Ie the normal assets endpoint are not available for media libraries, and some resources might not have assets at all. We also need the resources to be extendable with custom methods, though this can come later. Ie for media library we want a custom createAspect method

PR #1003, opened 2025-02-05
feat(listener): add support for includeAllVersions

bjoerge:
Adds support for passing the `includeAllVersions` param to `client.listen()`. Docs TBD

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -927,12 +927,7 @@ export interface ListenOptions {
   includePreviousRevision?: boolean
 
   /**
-   * Whether to include events for drafts and versions. As of API Version >= v2025-02-19, only events
-   * for published documents will be included by default.
-   * If you need events from drafts and versions, set this to `true`.
-   * Note: Keep in mind that additional document variants may be introduced in the future, so it's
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -920,6 +920,17 @@ export interface ListenOptions {
    */
   includePreviousRevision?: boolean
 
+  /**
+   * Whether to include events for drafts and versions. As of API Version >= v[TBD], only events
+   * for published documents will be included by default.
+   * If you need events from drafts and versions, set this to `true`.
+   * Note: Keep in mind that additional document variants may be introduced in the future, so it's
```
bjoerge: this assumes parity with `perspective=raw`, but given that the param is explicitly named includeAll*Versions*, I guess in theory we could introduce new variants in the future without including them behind this flag.

rneatherway: Right, that is possible. We'll have to consider any further developments in this area very carefully.

bjoerge:
This is now good to go. This will merely just forward the `includeAllVersions` param, which is marked as internal for now. The API docs can come later.

rneatherway (approved):
LGTM

PR #1005, opened 2025-02-06
fix: lift restriction on release perspective name

bjoerge:
We previously had a requirement on release perspectives that their names had to start with `r`. This restriction is no longer enforced on the backend, so it should not be required in the client either.

Note: the test failure is most likely unrelated and caused by flake. Have noticed it on PRs without code changes as well.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -28,6 +28,8 @@ function validateApiVersion(apiVersion: string) {
   }
 }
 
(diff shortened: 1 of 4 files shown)


stipsan (approved):
LGTM ❤

rneatherway (approved):

PR #1009, opened 2025-02-10
docs: add info about default perspective change

bjoerge:
Adds remarks about default perspective change effective from `v2025-02-19`

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,7 +501,7 @@ Perform a query using the given parameters (if any).
 
 ### Using perspectives
 
-The `perspective` option can be used to specify special filtering behavior for queries. The default value is `raw`, which means no special filtering is applied, while [`published`](#published) and [`previewDrafts`](#previewdrafts) can be used to optimize for specific use cases.
+The `perspective` option can be used to specify special filtering behavior for queries. The default value is [`published`](#published) if API version >= v2025-02-19, (if API version is <  v2025-02-19, [`raw`](#raw) is the default). Using the [`published`](#published) perspective will exclude drafts, versions and potential future document variants. Using the [`raw`](#raw) perspective means no special filtering is applied, so it's recommended to apply filtering client side that only includes the variants you want to process.
 
 #### `published`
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -501,7 +501,7 @@ Perform a query using the given parameters (if any).
 
 ### Using perspectives
 
-The `perspective` option can be used to specify special filtering behavior for queries. The default value is `raw`, which means no special filtering is applied, while [`published`](#published) and [`previewDrafts`](#previewdrafts) can be used to optimize for specific use cases.
+The `perspective` option can be used to specify special filtering behavior for queries. The default value is [`published`](#published) if API version >= v2025-02-19, (if API version is <  v2025-02-19, [`raw`](#raw) is the default). Using the [`published`](#published) perspective will exclude drafts, versions and potential future document variants. Using the [`raw`](#raw) perspective means no special filtering is applied, so it's recommended to apply filtering client side that only includes the variants you want to process.
```
markmichon:
```suggestion
Use the `perspective` option to filter queries from a specific viewpoint. The default value is [`published`](#published) if API version >= v2025-02-19, (if API version is < v2025-02-19, [`raw`](#raw) is the default). Using the [`published`](#published) perspective will exclude drafts, versions, and potential future document variants. Using the [`raw`](#raw) perspective returns all document variants,  so it's recommended to apply additional client side filtering to limit the variants you want to process.
```

Also, should we include perspective stack details here?

bjoerge: Thanks, great suggestions!

Just a minor comment, the original statement that said:
> (...) apply filtering client side that only includes the variants you want to process.

is not equivalent to 
> Use this [perspective=raw] client side to limit the variants you want to process.

Since perspective=raw gives you everything, we want to advice folks to filter out the known variants client side, so their logic won't be affected by potential variants added in the future.

bjoerge: Ah, and yes, we should include the perspective stack here as well. Do you have something already you can use, or do you want me to draft something?

markmichon: > Since perspective=raw gives you everything, we want to advice folks to filter out the known variants client side, so their logic won't be affected by potential variants added in the future.

Ah, bit of a misinterpretation on my end. Since we talk about perspective as a filtering technique, and it's on the client, it's a bit confusing to also mention filtering client side. I've updated the suggestion.

markmichon: > Ah, and yes, we should include the perspective stack here as well. Do you have something already you can use, or do you want me to draft something?

It won't show updates until the release docs go live, but https://www.sanity.io/docs/perspectives includes details on the concept. A "Learn more about using perspectives with Content Releases in the [perspective documentation](https://www.sanity.io/docs/perspectives)." should suffice.

I've been using something like `perspective: ['release-a', 'release-b', 'release-c']` in examples if you want to include an additional code example block just to reassure users that it accepts the different format.

bjoerge: Perfect! Yeah, I see the confusion. I've gradually shifted my mental model of perspectives to less of a filtering technique and more of a layering technique (e.g, perspective=drafts still gives you documents that are published)

bjoerge: Great! I'll draft something based on your links, thanks!

markmichon (approved):

stipsan (approved):

PR #1013, opened 2025-02-10
chore: actually solve edge runtime flake

stipsan:
Seems like a couple of tests have always either failed silently, or failed hard.
By always skipping over them (since nock doesn't work in the edge runtime) the flake is finally gone.


bjoerge (approved):
Thank you!

PR #1026, opened 2025-02-27
fix(docs): make dataset and project related types public

colepeters:
The SDK has a few hooks coming in that work with projects and datasets (https://github.com/sanity-io/sdk/pull/235). These hooks make use of some types from `sanity/client` which are currently marked `@internal` and thus aren't being surfaced in our Typedocs.

This PR just makes those types `@public` so we can surface them in our reference docs :)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,9 +1,9 @@
 // deno-lint-ignore-file no-empty-interface
 /* eslint-disable @typescript-eslint/no-empty-object-type */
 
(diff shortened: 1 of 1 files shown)


stipsan (approved):
Makes sense to me! I'll change your PR title prefix to `fix(docs):` so that release automation knows it should publish a new version. `docs:` are treated like `chore:` (unfortunately) 🙌

PR #1029, opened 2025-03-03
fix(docs): invalid changelog url

hermanwikner:
This pull request corrects the URL to the changelog for the default perspective change.


bjoerge (approved):
Huh, not sure what happened there. Good catch, thanks for taking the time to fix, @hermanwikner 🙏🏼

PR #1032, opened 2025-03-12
feat: releases client and document version actions

jordanl17:
[The necessary changes to use the client in studio](https://github.com/sanity-io/sanity/compare/main...refactor/sapp-2460-use-client-actions) (not currently up to date with latest changes)

Based on the server actions for [Versions](https://www.sanity.io/docs/http-actions#8c43ee58aa51) and [Releases](https://www.sanity.io/docs/http-actions#a102448a3a7a)

```
npm install @sanity/client@content-releases
```

> [!NOTE]
> Please don't critique/review the TSDoc comments or the error messages that get thrown yet - this will be followed up to the feature branch in a separate PR where docs will also review 

> [!IMPORTANT]
> This PR is into a feature branch `feat/sapp-243-releases-versions-client` - once this PR, and the docs one are merged to feature branch, then a separate PR into main will happen 

Options excluded in this approach:
* nesting versions as a separate client - eg `client.versions.create`, unless we also create a `documents` domain (`client.documents.create`) this feels incongruous
* Supporting a `version` in the existing `client.create` - although this feels somewhat intuitive, for the sake of keeping scope smaller, and getting some abstraction for versions into the client before making adjustments to existing methods

None of these feel like 1-way doors, so can be worked on as part of a wider client refactor in the future

# Reviewing
1. `eslint.config` updated to match a little more of what studio uses - so that unused imports and variables are appropriately warned and removed by the linter
2. `package.json` - adds `nanoid` which is also used in studio to generate the releaseIds. This is used so that `client.releases.create()` can be called without a releaseId, but will generate and return one. This is a very small package. This package only supports node >=18.0, so that has been updated in engine requirements
3. `src/types.ts` is mostly updated with copied types from [`packages/sanity/src/core/releases/store/types.ts`](https://github.com/sanity-io/sanity/blob/main/packages/sanity/src/core/releases/store/types.ts) - once this PR is releases then studio can be updated to use the type exports from client instead of defining its own
4. `src/util/createVersionId.ts` are some useful helpers: `generateReleaseId` is just a copy from studio; `getDocumentVersionId` is self explanatory; `deriveDocumentVersionId` is used so in version create and replace cases, where `document._id` or `publishedId` and `releaseId` are required, but if all are provided, they must match (otherwise an error is thrown)
5. `src/validators.ts` adds 3 new validators: `validateVersionIdMatch` used in `deriveDocumentVersionId`; `validateDocumentType` and `requireDocumentType`, following same pattern as is used to validate and require a documentId - `requireDocumentType` is used because [create and replace version operations require it](https://www.sanity.io/docs/http-actions#0511717e46fe)
6. `src/releases/ReleasesClient.ts` mostly just passed through to the actions data methods. The main callout being `.create` which accepts an optional release argument - if not provided then a `releaseId` is generated and a `releaseType` of `undecided` is used
7. `src/data/dataMethods.ts` exposes methods for the version operations - which are mostly pass-throughs to the action calls, with some additional document validation (on `_id` and `_type`)
8. `src/SanityClient.ts` where the chonk is - `getDocument` accepts an optional `options.releaseId` (which must match if the documentId is a version); `createVersion` and `replaceVersion` do validation on `document._id` and a provided `publishedId` and `releaseId` to check for consistency if all are provided. All methods use `drafts` as the fallback version if a `releaseId` is not provided

`ReleasesClient` and `SanityClient` do the same thing twice - for the observable and promise clients.

Plenty of tests added to cover all uses of these new methods, the validation on matching version IDs and `document._id`, overloaded signature uses etc.

# New APIs
### Get

```tsx
// returns the release system document
client.releases.get({
	releaseId: 'releaseId'
})
```

### Create

```tsx
client.action({
  actionType: 'sanity.action.release.create',
  releaseId: 'releaseId',
  // metadata not required
})

client.releases.create({
  // releaseId not required
  releaseId: 'releaseId',
  metadata: {
    releaseType: 'asap',
  },
})

// generates a releaseId and gives releaseType of 'undecided'
client.releases.create(actionOptions)
```

### Edit

```tsx
client.action({
  actionType: 'sanity.action.release.edit',
  releaseId: 'releaseId',
  patch: {
    set: {
      metadata: {
        title: 'title'
      }
    }
  }
})

client.releases.edit({
  releaseId: 'releaseId',
  patch: {
    set: {
      metadata: {
        title: 'title'
      }
    }
  }
})
```

> I haven't done it at this stage, but my preference would be that release.edit be more opinionated - you may only set `releaseType` to one of our values, you may only change/unset `intendedPublishAt` `title` `description` (and probably have a runtime check here for when TS is not used). Everything can be bypassed using the action directly.
> 
> 
> Would be interested if people feel similarly
> 

### Get Documents

```tsx
// returns all version documents with 'versions.releaseId.*' filter match
client.releases.getDocuments({
	releaseId: 'releaseId'
})
```

### Publish (similar to archive, unarchive, schedule, delete)

```tsx
client.action({
  actionType: 'sanity.action.release.publish',
  releaseId: 'releaseId',
})

client.releases.publish({releaseId: 'releaseId'}, actionOptions)
```

### Create

```tsx
client.action({
  actionType: 'sanity.action.document.version.create',
  publishedId: 'foo',
  document: {
    _id: 'versions.releaseId.foo'
    _type: 'type-required'
  })

client.createVersion({
    publishedId: 'foo',
    releaseId: 'releaseId',
    document: {
      // not required, must match if given
      _id: 'versions.releaseId.foo',
      _type: 'type-required'
    }
  },
  actionOptions
)

client.createVersion({
    document: {
      // required, if published and release IDs are not given
      _id: 'versions.releaseId.foo',
      _type: 'type-required'
    }
  },
  actionOptions
)

// create a draft - omitting `releaseId` infers `draft`
client.createVersion({
  publishedId: 'foo',
  document: {
    // not required, must match if given
    _id: 'drafts.foo', 
    _type: 'type-required'
  }
})
```

### Discard

```tsx
client.action({
  actionType: 'sanity.action.document.version.discard',
  versionId: 'versions.releaseId.foo'
})

client.discardVersion({
  releaseId: 'releaseId',
  publishedId: 'foo'
})

// discard a draft - omitting `releaseId` infers `draft`
client.discardVersion({
  publishedId: 'foo'
})
```

### Replace

```tsx
client.action({
  actionType: 'sanity.action.document.version.replace',
  document: {
    _id: 'versions.releaseId.foo',
    _type: 'type-required'
  }
})

client.replaceVersion({
  publishedId: 'foo',
  releaseId: 'releaseId',
  document: {
    // not required, must match if given
    _id: 'versions.releaseId.foo',
    _type: 'type-required'
  }
})

client.replaceVersion({
  document: {
    // required, if published and release IDs not given
    _id: 'versions.releaseId.foo',
    _type: 'type-required'
  }
})

// discard a draft - omitting `releaseId` infers `draft`
client.replaceVersion({
  publishedId: 'foo',
  document: {
    // not required, must match if given
    _id: 'drafts.foo',
    _type: 'type-required'
  }
})
```

### Unpublish

```tsx
client.action({
  actionType: 'sanity.action.document.version.unpublish',
  versionId: 'versions.releaseId.foo',
  publishedId: 'foo'
})

client.unpublishVersion({
  releaseId: 'releaseId',
  publishedId: 'foo'
})
```

> This seems nice since there's never going to be a deviation like there could be with using the action directly
> 

### GetDocument

```tsx
// 'versions.releaseId.foo' returned
client.getDocument('foo', {releaseId: 'releaseId'})
client.getDocument('versions.releaseId.foo')
client.getDocument('versions.releaseId.foo', {releaseId: 'releaseId'})

// throws error because of mismatch releaseId
client.getDocument('versions.releaseId.foo', {releaseId: 'bar'})

// throws error, version cannot be drafts
client.getDocument('foo', {releaseId: 'drafts'})

// 'drafts.foo' returned
client.getDocument('drafts.foo')

// throws error, because of mismatch version Id 
client.getDocument('drafts.foo', {releaseId: 'bar'})
```


jordanl17:
This pull request includes several updates to the ESLint configuration, package dependencies, and the addition of new functions and types for handling document versions and releases in the codebase. The most important changes are as follows:

### ESLint Configuration Updates:
* Added `eslint-plugin-unused-imports` to the ESLint configuration and disabled the `@typescript-eslint/no-unused-vars` rule in favor of the new plugin. (`eslint.config.mjs`: [[1]](diffhunk://#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R10) [[2]](diffhunk://#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R37) [[3]](diffhunk://#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R69-R72)

### Package Dependencies:
* Added `nanoid` and `eslint-plugin-unused-imports` to the `package.json` dependencies. (`package.json`: [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R123) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R146)

### Node Version Requirement:
* Updated the required Node.js version to 18 or higher in the `package.json`. (`package.json`: [package.jsonL164-R166](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L164-R166))

### New Functions for Document Versions and Releases:
* Added new functions in `src/data/dataMethods.ts` for handling document versions and releases, including `_getReleaseDocuments`, `_createVersion`, `_discardVersion`, `_replaceVersion`, and `_unpublishVersion`. (`src/data/dataMethods.ts`: [[1]](diffhunk://#diff-b09d77bf2334c027f21d5a51b58875a9d75d7df979ee5ae06fcdd8f8076fa7f7R172-R189) [[2]](diffhunk://#diff-b09d77bf2334c027f21d5a51b58875a9d75d7df979ee5ae06fcdd8f8076fa7f7R226-R245) [[3]](diffhunk://#diff-b09d77bf2334c027f21d5a51b58875a9d75d7df979ee5ae06fcdd8f8076fa7f7R269-R320)

### New Types and Interfaces:
* Introduced new types and interfaces in `src/types.ts` for handling release actions and document versions, such as `ReleaseAction`, `VersionAction`, `CreateReleaseAction`, and `CreateVersionAction`. (`src/types.ts`: [[1]](diffhunk://#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR555-R572) [[2]](diffhunk://#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR582-R711) [[3]](diffhunk://#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR1374-R1439)

src/SanityClient.ts
```diff
@@ -87,6 +91,7 @@ export class ObservableSanityClient {
     this.live = new LiveClient(this)
     this.projects = new ObservableProjectsClient(this, this.#httpRequest)
     this.users = new ObservableUsersClient(this, this.#httpRequest)
+    this.releases = new ObservableReleasesClient(this, this.#httpRequest)
```
jordanl17, line 101: Q: Do you prefer 
🎉 `client.releases.create` `client.releases.publish` etc
OR
❤️ `client.release.create`, `client.release.publish` etc

`releases` is better aligned with the current namespaces on the client (`users`, `projects`), however we've never introduced content lake specific namespaces. Given that our actions are `sanity.action.release.create`... perhaps there is argument for using singular

src/data/dataMethods.ts
```diff
@@ -165,6 +169,24 @@ export function _getDocuments<R extends Record<string, Any>>(
...
+  httpRequest: HttpRequest,
+  releaseId: string,
+  opts: BaseMutationOptions = {},
+): Observable<RawQueryResponse<SanityDocument<R>[]>> {
+  return _dataRequest(
+    client,
+    httpRequest,
+    'query',
+    {
+      query: `*[_id in path("versions.${releaseId}.*")]`,
```
jordanl17: I think we could also use 
```
query: `*[sanity::partOfRelease("${releaseId}")]`,
```
I'm thinking it could be a nice use, although it's only supported > 2025-02-19. Unsure if that is reason to avoid it

RitaDias: I would like to hear other people's thoughts on this as well.
I think we should make use of these things, even if it's only supported after a version imo that's why they are there to begin with and it's a nice bit of dog fooding on our part as well. The only exception is if the "partOfRelease" wasn't stable but I think it is?

Anyone has any opinions on this?

pedrobonamin: I think we should use the partOfRelease, and we should probably update the studio to use this function in the `useBundleDocuments`

src/data/dataMethods.ts
```diff
@@ -204,6 +226,26 @@ export function _createOrReplace<R extends Record<string, Any>>(
   return _create<R>(client, httpRequest, doc, 'createOrReplace', options)
 }
 
+/** @internal */
+export function _createVersion<R extends Record<string, Any>>(
+  client: ObservableSanityClient | SanityClient,
+  httpRequest: HttpRequest,
+  doc: IdentifiedSanityDocumentStub<R>,
+  publishedId: string,
```
pedrobonamin, line 264: It looks that we are missing here `ifExists: "fail" | "ignore"` option, and it should be part of the createVersionAction. right?

src/data/transaction.ts
```diff
@@ -76,6 +77,27 @@ export class BaseTransaction {
     return this._add({[op]: doc})
   }
 
+  createVersion<R extends Record<string, Any> = Record<string, Any>>({
```
pedrobonamin: Is there a reason why we are not adding the other methods to the transactions?

src/data/dataMethods.ts
jordanl17: - [x] use `partOfRelease` for returning docs in a `releaseId`

src/releases/ReleasesClient.ts
```diff
@@ -0,0 +1,478 @@
...
+    ] => {
+      if (
+        typeof releaseOrOptions === 'object' &&
+        releaseOrOptions !== null &&
+        ('releaseId' in releaseOrOptions || 'metadata' in releaseOrOptions)
+      ) {
+        return [releaseOrOptions.releaseId, releaseOrOptions.metadata || {}, maybeOptions]
+      }
+
+      return [undefined, {}, releaseOrOptions as BaseActionOptions]
```
pedrobonamin:
```suggestion
      return [ generateReleaseId(), {}, releaseOrOptions as BaseActionOptions]
```

And then below you don't need a `finalReleaseId`

jordanl17: Thank you!

- [x] Refactor `getArgs` to look tidier

src/data/transaction.ts
jordanl17: I'd actually forgotten I'd added this to transactions 🙃 
I think the other version methods should exist here, yeh. The release actions cannot be included in transactions as they need to be standalone (requirement from content lake I think to ensure stable locking of the version documents what moving between release states)

- [ ] Add all the version actions to transaction

src/data/dataMethods.ts
jordanl17: I see this is an option on the `document.create` action, but does it exist on the `document.version.create`?
I'm just going off what I see [here](https://www.sanity.io/docs/http-actions#0511717e46fe)

src/releases/ReleasesClient.ts
```diff
@@ -0,0 +1,478 @@
...
+  create(
+    release: {releaseId?: string; metadata?: Partial<ReleaseDocument['metadata']>},
+    options?: BaseActionOptions,
+  ): Observable<SingleActionResult & {releaseId: string; metadata: ReleaseDocument['metadata']}>
+  create(
+    releaseOrOptions?:
+      | {releaseId?: string; metadata?: Partial<ReleaseDocument['metadata']>}
+      | BaseActionOptions,
+    maybeOptions?: BaseActionOptions,
+  ): Observable<SingleActionResult & {releaseId: string; metadata: ReleaseDocument['metadata']}> {
```
pedrobonamin, line 64: Do we really need to support this 3 different type of parameters? I know it's nice to do it like so the consumer needs to do in some cases less work, but if we expose only 1 way to pass parameters maybe it will simplify a lot our implementation.

src/releases/ReleasesClient.ts
```diff
@@ -0,0 +1,478 @@
...
+  /**
+   * A schedule action queues a release for publishing at the given future time.
+   * The release is locked such that no documents in the release can be modified and
+   * no documents that it references can be deleted as this would make the publish fail.
+   * At the given time, the same logic as for the publish action is triggered.
+   *
+   * @param releaseId - The id of the release to schedule (with no prefix)
+   */
+  schedule(
+    {releaseId, publishAt}: {releaseId: string; publishAt: string},
```
pedrobonamin, line 166: Can we add a note to the `publishAt` saying it's a UTC timestamp and not just a string, also, should we validate that timestamp or is CL doing that for use?
What happens if I schedule in the past?

src/releases/ReleasesClient.ts
```diff
@@ -0,0 +1,478 @@
...
+      ...metadata,
+      releaseType: metadata.releaseType || 'undecided',
+    }
+
+    const createAction: CreateReleaseAction = {
+      actionType: 'sanity.action.release.create',
+      releaseId: finalReleaseId,
+      metadata: finalMetadata,
+    }
+
```
pedrobonamin: Can we have the `create` action as it's own abstraction and we reuse it here and in the observable client?

test/util/createVersionId.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      })
+    })
+  })
+
+  describe('getDocumentVersionId', () => {
+    it('returns a draft ID when no releaseId is provided', () => {
+      const publishedId = 'document123'
+      const versionId = getDocumentVersionId(publishedId)
+      expect(versionId).toBe('drafts.document123')
+    })
```
pedrobonamin, line 40: This feels like it could confuse some users, what if instead of returning a draft id we throw?

src/util/shareReplayLatest.ts
```diff
@@ -44,7 +44,7 @@ function _shareReplayLatest<T>(config: ShareReplayLatestConfig<T>): MonoTypeOper
     let latest: T | undefined
     let emitted = false
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    // eslint-disable-next-line unused-imports/no-unused-vars
```
pedrobonamin: Maybe this change is not needed?

src/SanityClient.ts
pedrobonamin: I think singular will be better here, all the actions added to the ReleasesClient is to do actions only in 1 release at a time.

test/util/createVersionId.test.ts
pedrobonamin: nvm. it's ok, I see that you use it as part of the createVersion function which accepts either a version id or nothing, which will create a draft. So you need this to be undefined

src/SanityClient.ts
```diff
@@ -900,9 +1096,35 @@ export class SanityClient {
...
+        throw new Error(
+          `The document ID (${id}) is already a version of ${documentIdReleaseId} release, but this does not match the provided \`options.releaseId\` (${options.releaseId})`,
+        )
+      }
+
+      return id
+    }
+
+    const docId = getDocId()
+
```
pedrobonamin: Could we extract this and reuse in the observable client?

src/types.ts
```diff
@@ -1244,6 +1392,72 @@ export interface ActionErrorItem {
...
+  /**
+   * typically
+   * `_.releases.<name>`
+   */
+  _id: string
+  /**
+   * typically
+   * `<name>`
+   */
+  name: string
```
pedrobonamin: Is this comment necessary?

pedrobonamin (changes requested):
Looking good, added a couple of comments.

src/data/dataMethods.ts
pedrobonamin: sorry! my bad, the I landed in the document. create https://www.sanity.io/docs/http-actions#b17bdb1336b2

src/data/transaction.ts
pedrobonamin: Let's confirm that requirement with CL, I think it would be nice to have a transactions in where you create a release and then attach certain documents to that release. 
Like:
create release summer.
create document home for summer

all in 1 transaction

src/releases/ReleasesClient.ts
jordanl17: This is only exposing 2 different ways, passing just an optional options arg, or passing release and optional options. The third is the implemented signature. I think this is quite nice allowing for `client.releases.create()` to be used directly. The tests also cover all these overloaded types

src/releases/ReleasesClient.ts
jordanl17: Yeh this is added in my follow up PR for adding the docs and comments

src/util/shareReplayLatest.ts
jordanl17: I've disabled `@typescript-eslint/no-unused-vars` and turned on `unused-imports/no-unused-vars` since the latter supports autofixing

src/types.ts
jordanl17: - [x] improve the `ReleaseDocument['name']` comment

src/SanityClient.ts
jordanl17: Yeh I can have a look at something like that

- [x] can the `getDocument` observable and async client share logic

src/util/shareReplayLatest.ts
pedrobonamin: Ohh ok! I didn't got any error reverting this locally, thats why I asked

src/data/dataMethods.ts
```diff
@@ -168,6 +172,24 @@ export function _getDocuments<R extends Record<string, Any>>(
...
+  httpRequest: HttpRequest,
+  releaseId: string,
+  opts: BaseMutationOptions = {},
+): Observable<RawQueryResponse<SanityDocument<R>[]>> {
+  return _dataRequest(
+    client,
+    httpRequest,
+    'query',
+    {
+      query: `*[sanity::partOfRelease("${releaseId}")]`,
```
juice49: Minor: It might be nicer to use a param instead of string concatenation.

src/releases/ReleasesClient.ts
jordanl17: - [x] reuse release create across observable and promise client

src/SanityClient.ts
jordanl17: Extracted this logic into `dataMethods._getDocument` so it is shared there now instead of across both the clients

src/data/dataMethods.ts
jordanl17: Done

package.json
```diff
@@ -161,6 +163,6 @@
   },
   "packageManager": "npm@11.3.0",
   "engines": {
-    "node": ">=14.18"
```
bjoerge: I think we tend to consider this a breaking change. (a bit of a shame that we haven't been keeping it up-to-date with node LTS)

test/dataMethods.test.ts
```diff
@@ -21,6 +22,98 @@ describe('dataMethods', async () => {
   )
   const getClient = (conf?: ClientConfig) => createClient({...clientConfig, ...(conf || {})})
 
+  const createMockResponse = (documents: any[]) =>
```
bjoerge: Not a huge deal, but would prefer if documents could be parameterized instead of typed as `any`:


```ts
const createMockResponse = <T>(documents: T[]) =>
  of({
    type: 'response',
    body: {documents},
  })
```

test/dataMethods.test.ts
```diff
@@ -21,6 +22,98 @@ describe('dataMethods', async () => {
...
+    of({
+      type: 'response',
+      body: {documents},
+    })
+
+  const createMockQueryResponse = (result: any) =>
```
bjoerge: similar to above, would be preferable if result was a type parameter

test/dataMethods.test.ts
```diff
@@ -21,6 +22,98 @@ describe('dataMethods', async () => {
...
+    of({
+      type: 'response',
+      body: {result},
+    })
+
+  const assertObservable = <T>(observable: any, assertion: (value: T) => void): Promise<void> => {
```
bjoerge:
```suggestion
  const assertObservable = <T>(observable: Observable<T>, assertion: (value: T) => void): Promise<void> => {
```

src/data/dataMethods.ts
```diff
@@ -168,6 +199,27 @@ export function _getDocuments<R extends Record<string, Any>>(
   )
 }
 
+/** @internal */
+export function _getReleaseDocuments<R extends Record<string, Any>>(
+  client: ObservableSanityClient | SanityClient,
+  httpRequest: HttpRequest,
+  releaseId: string,
```
bjoerge, line 206: nit: should this be _releaseDocumentId_ for extra clarity?

test/releasesClient.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+      const releaseId = 'nonexistent'
+
+      nock(TEST_PROJECT_HOST)
+        .get(`/v1/data/doc/${TEST_DATASET}/_.releases.${releaseId}`)
+        .reply(200, {
+          documents: [],
+        })
+
+      httpRequest.mockImplementationOnce(() => ({
+        subscribe: (subscriber: any) => {
```
bjoerge: would be great if this could be typed as well. Should be able to use the Subscriber type from rxjs

test/releasesClient.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+
+    test('forwards signal and tag options', async () => {
+      const abortController = new AbortController()
+      const options = {
+        signal: abortController.signal,
+        tag: 'releases.get',
+      }
+
+      httpRequest.mockImplementationOnce(() => ({
+        subscribe: (subscriber: any) => {
```
bjoerge: same as above re: Subscriber type

test/releasesClient.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+
+  describe('getDocuments()', () => {
+    test('retrieves documents for a release by ID', async () => {
+      const documents = [
+        {_id: 'doc1', _type: 'post', title: 'Document 1'},
+        {_id: 'doc2', _type: 'post', title: 'Document 2'},
+      ]
+
+      httpRequest.mockImplementationOnce(() => ({
+        subscribe: (subscriber: any) => {
```
bjoerge: Same as above re: Subscriber type

test/releasesClient.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+    })
+
+    test('forwards options to underlying method', async () => {
+      const options = {
+        tag: 'releases.documents',
+        signal: new AbortController().signal,
+      }
+
+      httpRequest.mockImplementationOnce(() => ({
+        subscribe: (subscriber: any) => {
```
bjoerge: Same as above re: Subscriber type

test/releasesClient.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+      )
+
+      await expect(promise).rejects.toThrow('Request aborted')
+      await expect(promise).rejects.toHaveProperty('name', 'AbortError')
+    })
+  })
+})
+
+describe('ObservableReleasesClient', () => {
+  let client: any
```
bjoerge: type this?

test/releasesClient.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+  let observableReleasesClient: ObservableReleasesClient
+  let httpRequest: any
```
bjoerge: and this?

src/data/dataMethods.ts
```diff
@@ -132,10 +137,36 @@ export function _getDocument<R extends Record<string, Any>>(
...
   id: string,
-  opts: {signal?: AbortSignal; tag?: string} = {},
+  opts: {signal?: AbortSignal; tag?: string; releaseId?: string} = {},
 ): Observable<SanityDocument<R> | undefined> {
+  const getDocId = () => {
+    if (!opts.releaseId) {
+      return id
+    }
+
+    const documentIdReleaseId = getVersionFromId(id)
```
bjoerge: would versionId be a better name here?

eslint.config.mjs
```diff
@@ -33,6 +34,7 @@ export default [
     plugins: {
       '@typescript-eslint': typescriptEslint,
       'simple-import-sort': simpleImportSort,
+      'unused-imports': unusedImports,
```
bjoerge: would prefer if the eslint changes could be done in a separate PR

src/data/dataMethods.ts
jordanl17: The param to use with `sanity::partOfRelease` is the release name (what we are referring to as the `releaseId`. So where the release doc ID is `_.releases.xyz`, we'd call `sanity::partOfRelease(xyz)`, in which case I tend to feel having the arg as `releaseId` is more meaningful. Do you agree?

bjoerge: ah! I misread then, thought this would be the document id.

src/data/dataMethods.ts
```diff
@@ -132,10 +137,36 @@ export function _getDocument<R extends Record<string, Any>>(
...
+  opts: {signal?: AbortSignal; tag?: string; releaseId?: string} = {},
 ): Observable<SanityDocument<R> | undefined> {
+  const getDocId = () => {
+    if (!opts.releaseId) {
+      return id
+    }
+
+    const versionId = getVersionFromId(id)
+    if (!versionId) {
+      if (isDraftId(id)) {
```
bjoerge, line 149: Why do we not want to support `getDocument('drafts.foo', {releaseId: 'xyz'})`, or `getDocuments('versions.abc.foo', {releaseId: 'xyz'})`? Do we think it will help surface common errors?

src/SanityClient.ts
```diff
@@ -1265,7 +1558,7 @@ export class SanityClient {
   ): Promise<SanityDocument<R>>
   /**
    * Perform mutation operations against the configured dataset.
-   * Returns a promise that resolves to an array of the mutated documents.
+   * Returns an observable that resolves to an array of the mutated documents.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/SanityClient.ts
```diff
@@ -1287,18 +1580,18 @@ export class SanityClient {
   ): Promise<SingleMutationResult>
   /**
    * Perform mutation operations against the configured dataset
-   * Returns a promise that resolves to a mutation result object containing the mutated document IDs.
+   * Returns an observable that resolves to a mutation result object containing the mutated document IDs.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/SanityClient.ts
```diff
@@ -1276,7 +1569,7 @@ export class SanityClient {
   ): Promise<SanityDocument<R>[]>
   /**
    * Perform mutation operations against the configured dataset
-   * Returns a promise that resolves to a mutation result object containing the document ID of the first mutated document.
+   * Returns an observable that resolves to a mutation result object containing the document ID of the first mutated document.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/SanityClient.ts
```diff
@@ -1252,9 +1454,100 @@ export class SanityClient {
...
+
+    return lastValueFrom(
+      dataMethods._unpublishVersion(this, this.#httpRequest, versionId, publishedId, options),
+    )
+  }
+
   /**
    * Perform mutation operations against the configured dataset
-   * Returns a promise that resolves to the first mutated document.
+   * Returns an observable that resolves to the first mutated document.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/SanityClient.ts
```diff
@@ -1287,18 +1580,18 @@ export class SanityClient {
...
    */
-  mutate<R extends Record<string, Any>>(
+  mutate<R extends Record<string, Any> = Record<string, Any>>(
     operations: Mutation<R>[] | Patch | Transaction,
     options: AllDocumentIdsMutationOptions,
   ): Promise<MultipleMutationResult>
   /**
    * Perform mutation operations against the configured dataset
-   * Returns a promise that resolves to the first mutated document.
+   * Returns an observable that resolves to the first mutated document.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/SanityClient.ts
```diff
@@ -1372,7 +1665,6 @@ export class SanityClient {
 
   /**
    * Perform action operations against the configured dataset
-   * Returns a promise that resolves to the transaction result
```
bjoerge: why was this removed?

src/SanityClient.ts
```diff
@@ -1252,9 +1454,100 @@ export class SanityClient {
...
+  }
+
+  /**
+   * Replaces an existing version document.
+   *
+   * At least one of the **version** or **published** documents must exist.
+   * If the `releaseId` is provided, the version will be stored under the corresponding release.
+   * Otherwise, the **draft version** of the document will be replaced.
+   *
+   * @returns an observable that resolves to the `transactionId`.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/SanityClient.ts
```diff
@@ -1252,9 +1454,100 @@ export class SanityClient {
...
+      dataMethods._replaceVersion<R>(this, this.#httpRequest, documentVersion, options),
+    )
+  }
+
+  /**
+   * Used to indicate when a document within a release should be unpublished when
+   * the release is published.
+   * Note that `publishedId` must currently exist.
+   *
+   * @returns an observable that resolves to the `transactionId`.
```
bjoerge: This doesn't seem accurate? The return type is Promise?

src/releases/ReleasesClient.ts
```diff
@@ -0,0 +1,423 @@
...
+  getDocuments(
+    {releaseId}: {releaseId: string},
+    options?: BaseMutationOptions,
+  ): Observable<RawQueryResponse<SanityDocument[]>> {
+    return _getReleaseDocuments(this.#client, this.#httpRequest, releaseId, options)
+  }
+}
+
+/** @internal */
```
bjoerge: This is marked as internal, but it's exposed on `client.releases`. Is that intentional?

src/releases/ReleasesClient.ts
```diff
@@ -0,0 +1,423 @@
...
+  ): Promise<SingleActionResult> {
+    const deleteAction: DeleteReleaseAction = {
+      actionType: 'sanity.action.release.delete',
+      releaseId,
+    }
+
+    return lastValueFrom(_action(this.#client, this.#httpRequest, deleteAction, options))
+  }
+
+  getDocuments(
```
bjoerge: I think if this should be named differently to better reflect that it goes against the `/query` endpoint and not the `/doc` endpoint, which offers different guarantees.


We already have `client.getDocument()` and `client.getDocuments()`, which both uses the [doc](https://www.sanity.io/docs/http-doc) endpoint, so it becomes easy to think that `releases.getDocuments()` does as well, and that it comes with the same guarantees (in terms of fresh data, etc.), but that's not the case, since it uses the query endpoint.

Currently queries are done via `client.fetch('...')`, so maybe better if this was named `releases.fetchDocuments()`? Or, `releases.queryDocuments()`?

(In retrospect maybe `client.fetch()` should've been named `client.query(...)`)

jordanl17:
This pull request introduces support for managing document versions and releases in the `SanityClient` and `ObservableSanityClient` classes. Key updates include new methods for creating, replacing, discarding, and unpublishing document versions, as well as enhanced functionality for handling release-specific document operations.

### New Features for Version and Release Management:

* **Support for Document Versions:**
  - Added methods `createVersion`, `replaceVersion`, `discardVersion`, and `unpublishVersion` to both `SanityClient` and `ObservableSanityClient` for managing document versions. These methods allow for creating new versions, replacing existing ones, discarding drafts or release-specific versions, and marking documents for unpublishing when a release is published. [[1]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fR455-R508) [[2]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fR627-R711) [[3]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fR1283-R1338) [[4]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fR1457-R1550)

* **Release-Specific Document Retrieval:**
  - Updated the `getDocument` method to support fetching documents by release ID, ensuring compatibility with release-specific versions. Added validation to handle mismatched or invalid release IDs. [[1]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fL222-R227) [[2]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fL903-R1049) [[3]](diffhunk://#diff-b09d77bf2334c027f21d5a51b58875a9d75d7df979ee5ae06fcdd8f8076fa7f7L135-R169)

### Dependency Updates:

* Added `nanoid` to `dependencies` in `package.json` for generating unique IDs, likely used in versioning or release management.

### Internal Enhancements:

* **Release Document Querying:**
  - Introduced a new `_getReleaseDocuments` function in `dataMethods.ts` to query all documents associated with a specific release.

* **Utility Functions for Versioning:**
  - Imported utility functions like `getVersionId`, `getVersionFromId`, and `deriveDocumentVersionId` to facilitate version ID generation and validation. [[1]](diffhunk://#diff-d181f4948a51cd3a09266a2f439e7285aea2a235ea2cb5e1dfb9a79951bcc78fR49) [[2]](diffhunk://#diff-b09d77bf2334c027f21d5a51b58875a9d75d7df979ee5ae06fcdd8f8076fa7f7R1)

These changes enhance the `SanityClient` API by providing robust tools for managing document versions and releases, enabling more granular control over content workflows.

src/data/dataMethods.ts
jordanl17: In these 2 cases the version of the document (`drafts` and `abc`) doesn't match the version implied by the `releaseId`. It's ambiguous to know if you mean to get `drafts.foo` or `versions.xyz.foo` in the first example. Is there a different flow you are thinking where this could apply?

src/releases/ReleasesClient.ts
jordanl17: Good point! I'm thinking of going with `fetchDocuments` here, to keep consistent with `client.fetch(query)` that already exists, is that ok for you?

src/data/dataMethods.ts
bjoerge: yeah, think you're right. I was looking at it from the perspective of knowing that I had a draft id and wanting to get it as a version and felt that I should be able to do
`getDocument('drafts.foo', {releaseId: 'xyz'})` without having to first convert `drafts.foo` to its published id. But agree it's ambiguous, and in most cases these IDs are opaque to whoever writes the code, so think it's better to lean on some strictness here. Also easy for us to lift the restriction later if we change our minds.

src/releases/ReleasesClient.ts
bjoerge: Works for me!

bjoerge (approved):
Good to merge from my side. Great work @jordanl17! 🚀

PR #1033, opened 2025-03-17
chore(deps): upgrade happy-dom, vitest, pkg-utils

rexxars:
happy-dom has a vulnerability that is creating noise when installing deps.
vitest needed to be bumped along with it.
pkg-utils mostly for vibes.


stipsan (approved):
Seems like `npm run test:browser` is failing?

PR #1036, opened 2025-03-18
fix: remove client side validation of api perspective

bjoerge:
This removes our client side validation of release ids passed in the perspective array. Although it provides a slight education opportunity, and might help with debugging, it's probably better to leave this validation to the server, if we want it at all.


jordanl17 (approved):
🚀

PR #1040, opened 2025-03-25
feat: add experimental support for resources

sgulseth:
TL;DR:
Add support for resource data requests

Changes:
* Adds new optional config parameter `~experimental_resource`
* Uses `~` to get lower priority in IDEs
* Is mainly to configure access to data endpoints for resources
* Will throw an error if used with mgmt APIs, also dataset assets aren't available as a resource yet.

### DIFF
--- a/src/assets/AssetsClient.ts
+++ b/src/assets/AssetsClient.ts
@@ -7,6 +7,7 @@ import type {
   Any,
   HttpRequest,
   HttpRequestEvent,
+  InitializedClientConfig,
   ResponseEvent,
   SanityAssetDocument,
   SanityImageAssetDocument,
@@ -149,8 +150,7 @@ function _upload(
     meta = ['none']
   }
 
-  const dataset = validators.hasDataset(client.config())
-  const assetEndpoint = assetType === 'image' ? 'images' : 'files'
+  const config = client.config()
   const options = optionsFromFile(opts, body)
   const {tag, label, title, description, creditLine, filename, source} = options
   const query: Any = {
@@ -166,15 +166,46 @@ function _upload(
     query.sourceName = source.name
     query.sourceUrl = source.url
   }
+
   return _requestObservable(client, httpRequest, {
     tag,
     method: 'POST',
     timeout: options.timeout || 0,
-    uri: `/assets/${assetEndpoint}/${dataset}`,
+    uri: buildAssetUploadUrl(config, assetType),
     headers: options.contentType ? {'Content-Type': options.contentType} : {},
     query,
     body,
   })
+}
+
+function buildAssetUploadUrl(config: InitializedClientConfig, assetType: 'image' | 'file'): string {
+  const assetTypeEndpoint = assetType === 'image' ? 'images' : 'files'
+
+  if (config['~experimental_resource']) {
+    const {type, id} = config['~experimental_resource']
+    switch (type) {
+      case 'dataset': {
+        throw new Error(
+          'Assets are not supported for dataset resources, yet. Configure the client with `{projectId: <projectId>, dataset: <datasetId>}` instead.',
+        )
+      }
+      case 'canvas': {
+        return `/canvases/${id}/assets/${assetTypeEndpoint}`
+      }
+      case 'media-library': {
+        return `/media-libraries/${id}/upload`
+      }
+      case 'dashboard': {
+        return `/dashboards/${id}/assets/${assetTypeEndpoint}`
+      }
+      default:
+        // @ts-expect-error - handle all supported resource types
+        throw new Error(`Unsupported resource type: ${type.toString()}`)
+    }
+  }
+
+  const dataset = validators.hasDataset(config)
+  return `assets/${assetTypeEndpoint}/${dataset}`
 }
 
 function optionsFromFile(opts: Record<string, Any>, file: Any) {
--- a/src/data/dataMethods.ts
+++ b/src/data/dataMethods.ts
@@ -17,6 +17,7 @@ import type {
   HttpRequest,
   HttpRequestEvent,
   IdentifiedSanityDocumentStub,
+  InitializedClientConfig,
   InitializedStegaConfig,
   MultipleActionResult,
   MultipleMutationResult,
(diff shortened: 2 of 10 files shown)


sgulseth:
Not sure who is best to review this change, only followed "recently xxx" - feel free to dismiss, or assign to someone else

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sanity/client",
-  "version": "6.28.4",
```
snorrees: Strip before merge

src/data/dataMethods.ts
```diff
@@ -367,6 +368,27 @@ export function _create<R extends Record<string, Any>>(
   return _dataRequest(client, httpRequest, 'mutate', {mutations: [mutation]}, opts)
 }
 
+const hasDataConfig = (client: SanityClient | ObservableSanityClient) =>
```
snorrees: `type Client = SanityClient | ObservableSanityClient` would make this quite a bit more readable I think

snorrees (dismissed):
Really thorough test-suite here, 👍 from me.

Only thing that gave me pause was the need maintain two switch cases on resource type in buildAssetUploadUrl and resourceDataBase. I see there is some differences, so trying to DRY it up will probably not improve the code, just mentioning it.

src/data/dataMethods.ts
```diff
@@ -367,11 +370,32 @@ export function _create<R extends Record<string, Any>>(
   return _dataRequest(client, httpRequest, 'mutate', {mutations: [mutation]}, opts)
 }
 
+const hasDataConfig = (client: Client) =>
```
snorrees, line 373: Nice! This is super nit, feel free to ignore: could I get a newline between each function?

snorrees (dismissed):

src/validators.ts
```diff
@@ -76,3 +76,34 @@ export const requestTag = (tag: string) => {
 
   return tag
 }
+
+export const resourceConfig = (config: InitializedClientConfig): void => {
+  if (!config['~experimental_resource']) {
+    throw new Error('`resource` must be provided to perform resource queries')
```
stipsan, line 82: Does it make sense to add a test for this in `client.test`?

src/validators.ts
```diff
@@ -76,3 +76,34 @@ export const requestTag = (tag: string) => {
...
+      return
+    }
+    case 'dashboard':
+    case 'media-library':
+    case 'canvas': {
+      return
+    }
+    default:
+      // @ts-expect-error - handle all supported resource types
+      throw new Error(`Unsupported resource type: ${type.toString()}`)
```
stipsan, line 101: Should we have test coverage here?

src/assets/AssetsClient.ts
```diff
@@ -166,17 +166,48 @@ function _upload(
...
+function buildAssetUploadUrl(config: InitializedClientConfig, assetType: 'image' | 'file'): string {
+  const assetTypeEndpoint = assetType === 'image' ? 'images' : 'files'
+
+  if (config['~experimental_resource']) {
+    const {type, id} = config['~experimental_resource']
+    switch (type) {
+      case 'dataset': {
+        throw new Error(
+          'Assets are not supported for dataset resources, yet. Configure the client with `{projectId: <projectId>, dataset: <datasetId>}` instead.',
+        )
```
stipsan, line 190: Should we cover this in tests?

src/assets/AssetsClient.ts
```diff
@@ -166,17 +166,48 @@ function _upload(
...
+      }
+      case 'canvas': {
+        return `/canvases/${id}/assets/${assetTypeEndpoint}`
+      }
+      case 'media-library': {
+        return `/media-libraries/${id}/upload`
```
stipsan, line 196: Should we cover this branch in the test suite?

src/assets/AssetsClient.ts
```diff
@@ -166,17 +166,48 @@ function _upload(
...
+      }
+      case 'dashboard': {
+        return `/dashboards/${id}/assets/${assetTypeEndpoint}`
+      }
+      default:
+        // @ts-expect-error - handle all supported resource types
+        throw new Error(`Unsupported resource type: ${type.toString()}`)
```
stipsan, line 203: This is also missing in test coverage

src/data/dataMethods.ts
```diff
@@ -547,3 +562,32 @@ function _createAbortError(signal?: AbortSignal) {
 
   return error
 }
+
+const resourceDataBase = (config: InitializedClientConfig): string => {
+  if (!config['~experimental_resource']) {
+    throw new Error('`resource` must be provided to perform resource queries')
```
stipsan, line 574: Should we have test coverage here?

src/data/dataMethods.ts
```diff
@@ -547,3 +562,32 @@ function _createAbortError(signal?: AbortSignal) {
...
+  }
+  const {type, id} = config['~experimental_resource']
+
+  switch (type) {
+    case 'dataset': {
+      const segments = id.split('.')
+      if (segments.length !== 2) {
+        throw new Error('Dataset ID must be in the format "project.dataset"')
```
stipsan, line 582: And here?

src/data/dataMethods.ts
```diff
@@ -547,3 +562,32 @@ function _createAbortError(signal?: AbortSignal) {
...
+    }
+    case 'media-library': {
+      return `/media-libraries/${id}`
+    }
+    case 'dashboard': {
+      return `/dashboards/${id}`
+    }
+    default:
+      // @ts-expect-error - handle all supported resource types
+      throw new Error(`Unsupported resource type: ${type.toString()}`)
```
stipsan, line 597: And here? 😇

stipsan (dismissed):
This is great! 🫶 

My other comments are callouts where the coverage report points to uncovered lines in the code you're adding. Take it as optional suggestions, and not as cheeky blockers ;)

src/data/dataMethods.ts
sgulseth: In theory should never reach this point as we have the same check when setting the config. We have:
```
test('throws if resource type is dataset and id has no dots', () => {
      expect(() =>
        createClient({'~experimental_resource': {type: 'dataset', id: 'abc123'}}),
      ).toThrow(/Dataset resource ID must be in the format "project.dataset"/)
    })```

src/data/dataMethods.ts
sgulseth: Snorre ~Eskeland~ Linter Brekke!

sgulseth:
@stipsan @snorrees  thanks! Added more tests cases - but ignored the ones we guard for earlier. Ie we don't have a separate test suite for AssetsClient.ts, but we can add that on the next pass

snorrees (approved):

bjoerge (approved):

PR #1041, opened 2025-03-25
feat: agent actions API integration (vX only)

snorrees:
Adds the following new functions:

* `client.agent.action.generate()`
* `client.agent.action.transform()`
* `client.agent.action.translate()`

The API these use are only available on `apiVersion: 'vX'` – so make sure to `client.withConfig({apiVersion: 'vX'})`


src/instruct/types.ts
```diff
@@ -0,0 +1,369 @@
...
+   * Root target path.
+   *
+   * Use this to have the instruction only affect a part of the document.
+   *
+   * To further control the behavior of individual paths under the root, use `include`, `exclude`, `types.include`
+   * and `types.exclude`.
+   *
+   * Example:
+   *
+   * `target: ['body', {_key: 'someKey'}, 'nestedObject']`
```
markmichon: Should this be `path` rather than `target`?

snorrees: Yep!

src/agent/actions/transform.ts
```diff
@@ -0,0 +1,135 @@
...
+  /** schemaId as reported by sanity deploy / sanity schema store */
+  schemaId: string
+
+  /**
+   * The source document the transformation will use as input.
+   */
+  documentId: string
+
+  /**
+   * The target document will first get content copied over from the source,
```
markmichon:
```suggestion
   * The source document's content is first copied to the target,
```
It's a little confusing here to say 'the target document' as it kind of implies that the target is being copied, rather than the source (`documentId`) above.

EoinFalconer (dismissed):

EoinFalconer (approved):
LGTM

PR #1043, opened 2025-04-08
fix: only allow non-nullable values in search params

azaxarov:
Sanity Client accepts both nullable and non-nullable search parameters. Due to the fact that our API rejects values that are `undefined`, it is best that we filter them out and thus only allow nun-nullable values.

### DIFF
--- a/src/data/encodeQueryString.ts
+++ b/src/data/encodeQueryString.ts
@@ -18,7 +18,7 @@ export const encodeQueryString = ({
 
   // Iterate params, the keys are prefixed with `$` and their values JSON stringified
   for (const [key, value] of Object.entries(params)) {
-    searchParams.append(`$${key}`, JSON.stringify(value))
+    if (value !== undefined) searchParams.append(`$${key}`, JSON.stringify(value))
   }
   // Options are passed as-is
   for (const [key, value] of Object.entries(opts)) {
--- a/test/encodeQueryString.test.ts
+++ b/test/encodeQueryString.test.ts
@@ -43,9 +43,7 @@ test('can encode null nested values', () => {
   )
 })
 
-// eslint-disable-next-line no-warning-comments
-// @TODO handled in https://github.com/sanity-io/client/pull/1043
-test.skip('skips encoding undefined params', () => {
+test('skips encoding undefined params', () => {
   const query = '*[defined(slug.current) && slug.current == $slug]'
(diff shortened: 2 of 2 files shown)


src/data/encodeQueryString.ts
```diff
@@ -18,7 +19,7 @@ export const encodeQueryString = ({
 
   // Iterate params, the keys are prefixed with `$` and their values JSON stringified
   for (const [key, value] of Object.entries(params)) {
-    searchParams.append(`$${key}`, JSON.stringify(value))
+    if (isDefined(value)) searchParams.append(`$${key}`, JSON.stringify(value))
```
RitaDias: I don't have a lot of context on the client so I will leave it to other folks to weight in, however, this might be considered a breaking change for other devs that might have been using the ability to send in nullable values.

stipsan (changes requested):
Thanks for the PR 👋 

Please add tests that demonstrate the fix. A great test for this is one that fails if we run it on `main`, while it passes on your PR ❤ 

I ran some tests and it seems like the case you're wanting to fix is:
```js
await client.fetch(`*[_type == "post" && slug.current == $slug]`, {slug: undefined})
```
Is that correct? I was surprised to see the client throwing an error here.

At the same time, this is valid and need to continue to pass (is this what you allude to @RitaDias?):
```js
await client.fetch(`*[_type == "post" && slug.current == $slug]`, {slug: null})
```

As a common pattern is to read params from an `URLSearchParams` instance:
```js
const url = new URL(location.href)
const slug = url.searchParams.get('slug')
const post = await client.fetch(
  `*[_type == "post" && defined(slug.current) && slug.current == $slug]`, 
  {slug} // string, or `null` if the search param doesn't exist
)
```
Thus we have to continue forwarding `null` 🙌

bjoerge:
@azaxarov whats the status on this one? Unassigning myself as reviewer, but feel free to add me back once it's ready for review again :)

mak-koda:
my team is waiting for this, what's the status?

src/data/encodeQueryString.ts
azaxarov: You are right, @RitaDias. The code now only prevents `undefined` values from being passed to search parameters.

stipsan (approved):
Thanks! 💖

PR #1058, opened 2025-04-28
chore: add unused-imports plugin and update rules

jordanl17:
`eslint-plugin-unused-imports` is a very small (4kb) dev dependency only.
It is used in SDK and in sanity studio.

The principle difference for it's inclusion is that it auto-fixes issues with unusued imports. This creates a consistent development experience across our largest of projects and repos


bjoerge (approved):
LGTM

PR #1065, opened 2025-05-06
fix: improve query parse errors

rexxars:
If you have an invalid GROQ query, it can be hard to actually understand what went wrong - and even that it is caused by a GROQ syntax error. As an example, the query `*[_type == "event]` leaves out a missing `"` and produces the following error:

```
Error: unexpected token "\"event]", expected expression
```

Where is this coming from? What was the query?

With this change, we produce more context on GROQ query parse errors:
```
GROQ query parse error: unexpected token "\\"event]", expected expression.

*[_type == "event
-----------^^^^^^^

Line: 1
Column: 11
Tag: some-tag
```

The tag is only included if there is one, obviously.

### DIFF
--- a/src/http/errors.ts
+++ b/src/http/errors.ts
@@ -1,4 +1,8 @@
-import type {ActionError, Any, ErrorProps, MutationError} from '../types'
+import type {HttpContext} from 'get-it'
+
+import type {ActionError, Any, ErrorProps, MutationError, QueryParseError} from '../types'
+import {codeFrame} from '../util/codeFrame'
+import {isRecord} from '../util/isRecord'
 
 const MAX_ITEMS_IN_ERROR_MESSAGE = 5
 
@@ -9,8 +13,8 @@ export class ClientError extends Error {
   responseBody: ErrorProps['responseBody']
   details: ErrorProps['details']
 
-  constructor(res: Any) {
-    const props = extractErrorProps(res)
+  constructor(res: Any, context?: HttpContext) {
+    const props = extractErrorProps(res, context)
     super(props.message)
     Object.assign(this, props)
   }
@@ -30,7 +34,7 @@ export class ServerError extends Error {
   }
 }
 
-function extractErrorProps(res: Any): ErrorProps {
+function extractErrorProps(res: Any, context?: HttpContext): ErrorProps {
   const body = res.body
   const props = {
     response: res,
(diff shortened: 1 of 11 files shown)


src/http/errors.ts
```diff
@@ -96,6 +117,25 @@ function isPlainObject(obj: Any): obj is Record<string, unknown> {
   return typeof obj === 'object' && obj !== null && !Array.isArray(obj)
 }
 
+export function groqParseMessage(res: QueryParseError, tag?: string | null) {
+  const {query, start, end, description} = res.error
+  const lineStart = query.slice(0, start).lastIndexOf('\n') + 1
+  const lineNumber = (query.slice(0, lineStart).match(/\n/g) || []).length + 1
+  const line = query.slice(lineStart, query.indexOf('\n', lineStart))
```
atombender: These calculations may be slightly off if there is no `\n` in the string, since `lastIndexOf()` would return -1.

Note that `start` and `end` might point to a broader region across multiple lines.

atombender:
Nice to see our structured errors actually being put to use. Regarding rendering the fragment, have you seen https://babeljs.io/docs/babel-code-frame?

rexxars:
> Regarding rendering the fragment, have you seen https://babeljs.io/docs/babel-code-frame?

I had not! Unfortunately it adds [quite some weight](https://bundlephobia.com/package/@babel/code-frame@7.27.1) to the bundle. Maybe we can revisit if a new version comes out marked with no side effects and proper tree shaking!

atombender:
Bundlephobia. 😬  Yeah, I was afraid of that…

rexxars:
Had a look at the source code, and it doesn't seem too bad - I can just inline it and give credits to the Babel team for it. With color coding removed and configurability reduced, we're down to about ~650 bytes added.

SimeonGriggs:
Love seeing this!

Two small notes:

This wording feels 🤏 clunky
> GROQ query parse error:

My suggestion
> Error, invalid GROQ query:

2. If the query is contained is in a variable (eg `EVENTS_QUERY`) would we be able to print the name of the variable? Or is this asking too much at run time :)

bjoerge (dismissed):
Love it!

rexxars:
>     2. If the query is contained is in a variable (eg `EVENTS_QUERY`) would we be able to print the name of the variable? Or is this asking too much at run time :)

I don't think there is any way for us to do this at runtime, no. The client just receives a string - it doesn't know if it came from a variable or it was statically declared.

bjoerge (dismissed):
Really great! 🚀

rexxars:
Apologies for review requests, had some minor lint issues.

bjoerge (approved):

PR #1075, opened 2025-05-09
chore: improvements to CI with PR title validation and esm compatibility check

jordanl17:
This introduces 2 improvements to the CI build steps to address 2 recent issues:
1. Adds in a validation that PR titles are conventional commit messages - this was found by a previous issue of a merged PR that did not trigger a new release-please version release
2. Adding a check that all production dependencies support CJS imports - recently a version of nanoid was added that does not support CJS, only ESM. This could have been spotted and prevented had this check already existed

Testing:
1. Since validate workflow uses the PR target as the source of the file, this cannot be fully tested until this PR is merged, however manual testing of using only `pull_request` as the trigger showed that the title does correctly fail the build
2. When using the version of nanoid that only supports ESM the follow was seen:
Screenshot 2025-05-10 at 00 57 48

### DIFF
--- a/.github/workflows/validate-pr-title.yml
+++ b/.github/workflows/validate-pr-title.yml
@@ -1,0 +1,21 @@
+name: 'Validate PR title'
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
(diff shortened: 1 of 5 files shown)


.github/workflows/validate-pr-title.yml
jordanl17, line 1: This is just a copy from the monorepo

juice49 (approved):
Seems very sensible.

PR #1076, opened 2025-05-09
fix: downgrade `nanoid` to ensure compatibility with node < 22

juice49:
The currently installed `nanoid` version (`^5.1.5`) cannot be imported by node < 22. This causes a breaking error when executing `@sanity/cli` on such a node version.

This branch downgrades to `nanoid@^3.3.11`, which is consistent with the version used by the Sanity monorepo, and can successfully be imported by node 22.


jordanl17 (approved):
Absolute saviour

PR #1084, opened 2025-05-20
viewsqueryclient fetch support

danielgroves:


src/views/index.ts
```diff
@@ -30,16 +45,100 @@ export class ViewClient {
...
+    query: G,
+    params: Q extends QueryWithoutParams ? QueryWithoutParams : Q,
+    options: UnfilteredResponseWithoutQuery,
+  ): Promise<RawQuerylessQueryResponse<ClientReturn<G, R>>>
   fetch<R, Q, const G extends string>(
     viewId: string,
     query: G,
     params?: Q,
-    options?: ViewsQueryOptions,
+    options?: QueryOptions,
```
danielgroves, line 141: This should probably be changed back

src/views/index.ts
```diff
@@ -30,11 +45,60 @@ export class ViewClient {
...
     this.#config = config
     this.#httpRequest = httpRequest
+
+    this.observable = new ObservableViewClient(httpRequest, config)
+  }
+
+  /**
+   * Clone the client with a new (partial) configuration.
+   *
+   * @param viewId - ID of the view to query
```
mwain: This seems wrong

mwain (approved):

sgulseth (commented):
Not directly related to this pr, but the resource type should be `view` and not `views`. same form as we have for the other resources media-library, canvas, dashboard, dataset.

This doesn't go into main, so I guess it's not meant to be stable, but would be nice with at least one test to test wanted behavior :p 
Ie are you meant to do: `createClient({
apiVersion: 'v2025-02-19'}).views.fetch('myId')`? The test in the parent only expose a separate client, so are we providing two clients?

danielgroves:
> are you meant to do: `createClient({ apiVersion: 'v2025-02-19'}).views.fetch('myId')`? The test in the parent only expose a separate client, so are we providing two clients?

We're providing a seperate client, at least for now. Views sit outside the context of a project, so although they will pull data _from_ projects to materialise, the actual view is a resource at the org level unlike the main SanityClient. We might change this later, but for now we're keen to keep it seperate.

PR #1086, opened 2025-05-21
feat(views): implement dev mode

danielgroves:
Add dev mode to views client:

Outbound requests need to look something like this:
```bash
POST /v2025-01-01/views/my-view-id/emulate?returnQuery=false HTTP/1.1
Host: apicdn.sanity.url
Content-Type: application/json
Accept: application/json

{
  "query": "*[_type == \"article\" && category == $cat]",
  "params": {
    "cat": "technology"
  },
  "connections": [
    {
      "query": "*[_type == \"document\"]",
      "resourceType": "dataset",
      "resourceId": "project123.dataset456"
    }
  ]
}
```

We make the request by calling the views client like this:

```js
import { createViewClient, ViewResourceType } from '@sanity/client/views'

const client = createViewClient({
  apiVersion: '2025-01-01',
  viewOverrides: [
    {
      id: 'my-view-id',
      connections: [
        {
          query: '*[_type == "document"]',
          resourceType: ViewResourceType.Dataset,
          resourceId: 'project123.dataset456'
        }
      ]
    }
  ]
})

// This automatically uses the emulate endpoint because dataset connections are detected
const result = await client.fetch(
  'my-view-id', 
  '*[_type == "article" && category == $cat]',
  { cat: 'technology' }
)
```

### DIFF
--- a/src/data/dataMethods.ts
+++ b/src/data/dataMethods.ts
@@ -90,12 +90,19 @@ export function _fetch<R, Q>(
   const mapResponse =
     options.filterResponse === false ? (res: Any) => res : (res: Any) => res.result
 
-  const {cache, next, ...opts} = {
+  const {cache, next, useEmulate, connections, ...opts} = {
     // Opt out of setting a `signal` on an internal `fetch` if one isn't provided.
     // This is necessary in React Server Components to avoid opting out of Request Memoization.
     useAbortSignal: typeof options.signal !== 'undefined',
     // Set `resultSourceMap' when stega is enabled, as it's required for encoding.
     resultSourceMap: stega.enabled ? 'withKeyArraySelector' : options.resultSourceMap,
+
+    // Only use emulate if explicitly asked for
+    useEmulate: false,
+
+    // Having connections is a special case for views
+    connections: undefined,
+
     ...options,
     // Default to not returning the query, unless `filterResponse` is `false`,
     // or `returnQuery` is explicitly set. `true` is the default in Content Lake, so skip if truthy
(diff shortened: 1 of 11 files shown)


src/data/dataMethods.ts
```diff
@@ -439,7 +458,7 @@ export function _dataRequest(
     resultSourceMap: options.resultSourceMap,
     lastLiveEventId: Array.isArray(lastLiveEventId) ? lastLiveEventId[0] : lastLiveEventId,
     cacheMode: cacheMode,
-    canUseCdn: isQuery,
+    canUseCdn: isQuery || isEmulate,
```
mwain: dont think we can use cdn when emulation mode

src/types.ts
```diff
@@ -1228,11 +1229,45 @@ export interface UnfilteredResponseWithoutQuery extends ResponseQueryOptions {
   returnQuery: false
 }
 
+/** @public */
+export enum ViewResourceType {
```
mwain, line 1233: this seems like a new concept to expose, is there any other way?

mwain (changes requested):

src/types.ts
danielgroves: Just tyring to keep the conventions in the right direction for consistency everywhere.

The existing client makes assumptions about the resoure type, which we could do, but in my opinion with a new client we shouldn't be afraid to move in the target direction. 

```js
sanityClient({
  projectId: xxxx,
  dataset: xxxx
})
```

We could follow that here too, but to me that just feels like we're intentionally adding something to fix later.

src/data/dataMethods.ts
danielgroves: Ah gotcha, I'll address this.

mwain (approved):

PR #1089, opened 2025-05-26
feat: new agent action parameter forcePublishedWrite – agent actions never write to published doc by default

snorrees:
**API BREAKING CHANGE**: The changes in this PR are not breaking for the _client api_, but the backend will start behaving differently once this has been released. Since agent actions require `/vX , this is new behavior will be forced upon all users of agent actions, regardless of client version. Using this release (or later) will be required to opt out of the new behavior.

## Agent actions never write to published documents by default

As of this going out, by default:
- agent.actions will NEVER mutate a published doc – it will use a drafts id if a published id is provided
- operations on drafts will get initial state from published doc, if no draft exists
- Callers should check the _id in the response as needed; when a `publishedId` is provided a `drafts.publishedId` will be returned
- To opt out of the default behavior, pass `forcePublishedWrite: true`
- Documents with `liveEdit: true` schema definition will behave as if `forcePublishedWrite: true`, and also accept drafts IDs in requests

 ### forcePublishedWrite: false (default)
 By default, agent actions will never write to a published document.

 Instead, they will force the use of a draft ID ("drafts.some-id") instead of the published ID ("some-id"),
 even when a published ID is provided.

 Actions will use state from an existing draft if it exists,
 or use the published document to create a draft, if no draft exists.

 Successful responses contain the _id that was mutated by the action.

 ### forcePublishedWrite: true

 When forcePublishedWrite: true an agent action will write to the exact id provided.
 The action will also not fallback to published state for draft ids.

 ### Versioned ids (releases)

 When an ID on the form "versions.<release>.some-id" is provided, agent actions will
 always behave as if `forcePublishedWrite: true`.
 That is, only the exact document state of the id provided is considered and mutated.

### DIFF
--- a/src/agent/actions/commonTypes.ts
+++ b/src/agent/actions/commonTypes.ts
@@ -235,6 +235,30 @@ export interface AgentActionSchema {
   schemaId: string
 
   /**
+   * ### forcePublishedWrite: false (default)
+   * By default, agent actions will never write to a published document.
+   *
+   * Instead, they will force the use of a draft ID ("drafts.some-id") instead of the published ID ("some-id"),
+   * even when a published ID is provided.
+   *
+   * Actions will use state from an existing draft if it exists,
+   * or use the published document to create a draft, if no draft exists.
+   *
+   * On successful responses contains the _id that was actually mutated by the action.
+   *
+   * ### forcePublishedWrite: true
+   *
+   * When forcePublishedWrite: true an agent action will write to the exact id provided.
+   * The action will also not fallback to published state for draft ids.
+   *
+   * ### Versioned ids (releases)
+   * When an ID on the form "versions.<release>.some-id" behave is provided, agent actions will
+   * always behave as if forcePublishedWrite: true. That is, only the exact document state of the id provided
+   * is considered.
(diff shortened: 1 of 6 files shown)


src/agent/actions/commonTypes.ts
```diff
@@ -234,6 +234,33 @@ export interface AgentActionSchema {
...
+   * ### forcePublishedWrite: false (default)
+   * By default, agent actions will never write to a published document.
+   *
+   * Instead, they will force the use of a draft ID ("drafts.some-id") instead of the published ID ("some-id"),
+   * even when a published ID is provided.
+   *
+   * Actions will use state from an existing draft if it exists,
+   * or use the published document to create a draft, if no draft exists.
+   *
+   * Successful responses contains the _id that was mutated by the action.
```
markmichon:
```suggestion
   * Successful responses contain the _id that was mutated by the action.
```

markmichon (approved):
left the smallest of suggestions.

robinpyon (approved):

mogenshoj (approved):

PR #1095, opened 2025-05-30
refactor: add typedoc support + cleanup

kmelve:
This PR introduces automated API documentation generation and upload for the `@sanity/client` package. It adds a new GitHub Actions workflow, configures TypeDoc, and updates the project to support documentation publishing on pushes to `main` and new releases.

---

#### 1. **GitHub Actions Workflow**
- **New workflow:** `.github/workflows/upload-docs.yml`
  - Triggers on pushes to `main` and on new releases.
  - Installs dependencies, builds docs using TypeDoc, and uploads them using the `sanity-io/reference-api-typedoc` action.
  - Uses a `SANITY_DOCS_API_TOKEN` secret for authentication.

#### 2. **TypeDoc Integration**
- **New config:** `typedoc.json`
  - Documents `src/types.ts` and `src/SanityClient.ts`.
  - Excludes test and declaration files, private/protected/internal members.
  - Outputs to the `docs` directory.

- **New script:** `docs` in `package.json`
  - Runs TypeDoc to generate JSON output for documentation.

- **New dev dependency:** `typedoc@^0.28.3` added to `package.json` and `package-lock.json`.

#### 3. **Project Structure & Ignore Rules**
- **.gitignore:** Now ignores the `docs` directory to prevent generated docs from being committed.

#### 4. **API & Type Improvements**
- Several types and interfaces in `src/types.ts` and related files are now exported as `public` for better TypeDoc visibility.
- Minor JSDoc and parameter name improvements for clarity in generated docs.
- Test type imports updated for consistency.

#### 5. **Documentation Improvements**
- Improved and clarified JSDoc comments for several public APIs, especially around release creation and versioning.

---

### Reviewer Notes

- **No breaking changes** to runtime code; all changes are additive or documentation-related.
- Some `@internal` types has been made `@public` since they are referenced in public types
- The workflow is non-intrusive and only runs on documentation-related triggers.
- TypeDoc config and script are designed to be minimal and non-disruptive to existing developer workflows.

---

#### How to test

- Run `npm run docs` locally to generate docs and verify output in the `docs` directory.
- You can easily serve the generated docs with `npx http-server -c-1 docs` 
- On merge to `main` or a new release, the workflow should automatically build and upload docs.

---

Let me know if you have any questions or suggestions!


src/types.ts
```diff
@@ -1606,42 +1608,81 @@ export type LiveEvent =
...
   GenerateTargetDocument,
   GenerateTargetInclude,
 } from './agent/actions/generate'
+export type {
+  GenerateAsyncInstruction,
+  GenerateExistingDocumentRequest,
+  GenerateRequestBase,
+  GenerateSyncInstruction,
+  GenerateTargetDocumentRequest,
+} from './agent/actions/generate'
```
snorrees: Non of these are intended to be exported, they are just assembly types used to arrive at `GenerateInstruction`. 
We should keep the export namespace as small as possible, these are just noise imo.

You are also exporting `from './agent/actions/generate'` twice

src/types.ts
```diff
@@ -1606,42 +1608,81 @@ export type LiveEvent =
...
 export type {PatchDocument, PatchOperation, PatchTarget} from './agent/actions/patch'
-export type {PromptRequest} from './agent/actions/prompt'
+export type {
+  PatchDocumentAsync,
+  PatchDocumentSync,
+  PatchExistingDocumentRequest,
+  PatchRequestBase,
+  PatchTargetDocumentRequest,
+} from './agent/actions/patch'
```
snorrees: Same here, internal types, should not be exported. And `from './agent/actions/patch'` twice.

src/types.ts
```diff
@@ -1606,42 +1608,81 @@ export type LiveEvent =
...
+  TransformRequestBase,
   TransformTarget,
   TransformTargetDocument,
   TransformTargetInclude,
 } from './agent/actions/transform'
 export type {
   TranslateDocument,
+  TranslateDocumentAsync,
+  TranslateDocumentSync,
+  TranslateLanguage,
```
snorrees: Keep `TranslateLanguage`, remove the others

src/types.ts
```diff
@@ -1606,42 +1608,81 @@ export type LiveEvent =
...
+export type {
+  PatchDocumentAsync,
+  PatchDocumentSync,
+  PatchExistingDocumentRequest,
+  PatchRequestBase,
+  PatchTargetDocumentRequest,
+} from './agent/actions/patch'
+export type {
+  PromptJsonResponse,
+  PromptRequest,
```
snorrees: Keep `PromptRequest`, remove the others

src/agent/actions/generate.ts
```diff
@@ -249,7 +249,7 @@ export type GenerateTargetDocument<T extends Record<string, Any> = Record<string
  * Instruction for an existing document.
  * @beta
  */
-interface GenerateExistingDocumentRequest {
+export interface GenerateExistingDocumentRequest {
```
snorrees: Is this really necessary? See comment below about assembly types. What are we hoping to achive with these exports?

snorrees (changes requested):

src/SanityClient.ts
jordanl17, line 1: There are other instances of `params` being used in `SanityClient` where it seems we'd now prefer `args`, eg line 703

src/types.ts
```diff
@@ -1606,42 +1608,81 @@ export type LiveEvent =
   | LiveEventGoAway
 
 /** @public */
-export interface SanityQueries {}
+export type FooResult = {
```
jordanl17: Q: Where does this get used in our exposed types?

src/types.ts
kmelve: Got it! I just exported everything that the typedoc tooling was complaining about. I can remove this, but it does mean that there will be some "dead ends" in the library reference docs, but maybe that's fine.

src/types.ts
kmelve: ack

src/SanityClient.ts
kmelve: I see! I didn't want to touch the actual source code, but needed to make the typedoc annotation param name map to what was actually used in the function definition. If that makes sense.

src/types.ts
kmelve: I wasn't sure about this, it seems like it was added to some kind of test code.

src/types.ts
snorrees: If we lose parameters in the docs tools (if fields cannot render with docs), then you have to obviously have to do what you have to do. I dont know how this all fits together, my point is that we should ideally only export what the user would directly use.

src/types.ts
snorrees: This smells like test code/types leaking into our production pacage.

Without knowing exactly how this works, I would be fairly certain to handwave that `interface SanityQueries {}` is very intentionally an empty interface which userland ts interface module merges groq queries into (typegen). 

What you have done here is take the demands of the client test suit and put it into prod. Revert for sure.

rexxars:
Belive this is superseded by #1273 - reopen if not

PR #1107, opened 2025-06-17
feat: add ignoreExperimentalApiWarning configuration option

kmelve:
## Summary

- Adds a flexible `ignoreWarnings` client configuration option to filter server warnings
- Replaces the specific `ignoreExperimentalApiWarning` initially pushed to branch with a generic pattern-matching system based on feedback
- Includes exported constants for common warning patterns with comprehensive documentation

## Problem

When using experimental API versions (`apiVersion: "vX"`), the client shows warnings like:
> "This is an experimental API version, which will change without warning and may have serious bugs."

The original approach used a specific boolean flag, but feedback indicated this should be more generic to handle any server warning, not just experimental API warnings.

## Solution

Implemented a flexible `ignoreWarnings` option that supports:

- **String patterns** - Exact substring matching
- **Regex patterns** - Advanced pattern matching with flags  
- **Arrays** - Mix and match multiple patterns
- **Exported constants** - For common warning patterns

## Usage Examples

```typescript
// String matching
const client = createClient({
  projectId: 'abc123',
  apiVersion: 'vX',
  ignoreWarnings: 'experimental API version'
})

// Regex pattern
const client = createClient({
  projectId: 'abc123', 
  apiVersion: 'vX',
  ignoreWarnings: /experimental.*version/i
})

// Multiple patterns
const client = createClient({
  projectId: 'abc123',
  apiVersion: 'vX', 
  ignoreWarnings: [
    /experimental/i,
    'rate limit',
    /deprecated/
  ]
})

// Using exported constants
import { createClient, EXPERIMENTAL_API_WARNING_PATTERN } from '@sanity/client'

const client = createClient({
  projectId: 'abc123',
  apiVersion: 'vX',
  ignoreWarnings: EXPERIMENTAL_API_WARNING_PATTERN
})
```

## API Design

**New configuration option:**
```typescript
interface ClientConfig {
  ignoreWarnings?: string | RegExp | Array<string | RegExp>
}
```

**Exported constants:**
```typescript
export const EXPERIMENTAL_API_WARNING = 'This is an experimental API version'
export const EXPERIMENTAL_API_WARNING_PATTERN = /This is an experimental API version/
```

## Changes

- **src/types.ts**: 
  - Added `ignoreWarnings` option to `ClientConfig` interface
  - Added `EXPERIMENTAL_API_WARNING` and `EXPERIMENTAL_API_WARNING_PATTERN` constants
  - Comprehensive TypeDoc documentation with usage examples
- **src/http/request.ts**: 
  - Implemented generic pattern matching in `printWarnings()` 
  - Supports string, regex, and array filtering
- **src/defineCreateClient.ts**: 
  - Updated to pass `ignoreWarnings` to HTTP middleware
- **test/warnings.test.ts**: 
  - Comprehensive test coverage for all pattern types
  - Tests for exported constants and array combinations

## Benefits

- **Generic**: Works for any server warning, not just experimental API
- **Flexible**: Supports exact strings, regex patterns, and combinations
- **Future-proof**: Can handle new warning types without code changes  
- **User-friendly**: Exported constants with clear documentation
- **Maintainable**: Single, clear warning filtering mechanism

## Test Plan

- [x] All existing tests pass
- [x] New tests verify string pattern matching
- [x] New tests verify regex pattern matching
- [x] New tests verify array combinations  
- [x] New tests verify exported constants work correctly
- [x] Build and lint checks pass
- [x] TypeDoc documentation renders properly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -22,7 +22,7 @@ function printWarnings(config: {ignoreWarnings?: string | RegExp | Array<string 
 
   // Helper function to check if a warning should be ignored
   const shouldIgnoreWarning = (message: string): boolean => {
-    if (!config.ignoreWarnings) return false
+    if (config.ignoreWarnings === undefined) return false
 
     const patterns = Array.isArray(config.ignoreWarnings)
       ? config.ignoreWarnings
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -17,7 +17,7 @@ const httpError = {
   },
 }
 
-function printWarnings() {
+function printWarnings(config: {ignoreExperimentalApiWarning?: boolean} = {}) {
   const seen: Record<string, boolean> = {}
   return {
     onResponse: (res: Any) => {
@@ -25,6 +25,15 @@ function printWarnings() {
       const warnings = Array.isArray(warn) ? warn : [warn]
       for (const msg of warnings) {
         if (!msg || seen[msg]) continue
+
+        // Skip experimental API warnings if configured to ignore them
+        if (
+          config.ignoreExperimentalApiWarning &&
+          msg.includes('This is an experimental API version')
+        ) {
+          continue
+        }
+
         seen[msg] = true
         console.warn(msg) // eslint-disable-line no-console
       }
@@ -34,11 +43,14 @@ function printWarnings() {
 }
 
 /** @internal */
-export function defineHttpRequest(envMiddleware: Middlewares): Requester {
+export function defineHttpRequest(
+  envMiddleware: Middlewares,
+  config: {ignoreExperimentalApiWarning?: boolean} = {},
+): Requester {
   return getIt([
     retry({shouldRetry}),
     ...envMiddleware,
-    printWarnings(),
+    printWarnings(config),
     jsonRequest(),
     jsonResponse(),
     progress(),
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -17,8 +17,27 @@ const httpError = {
   },
 }
 
-function printWarnings(config: {ignoreExperimentalApiWarning?: boolean} = {}) {
+function printWarnings(config: {ignoreWarnings?: string | RegExp | Array<string | RegExp>} = {}) {
   const seen: Record<string, boolean> = {}
+
+  // Helper function to check if a warning should be ignored
+  const shouldIgnoreWarning = (message: string): boolean => {
+    if (!config.ignoreWarnings) return false
+
+    const patterns = Array.isArray(config.ignoreWarnings)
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -1660,3 +1661,55 @@ export type {
...
+ *     EXPERIMENTAL_API_WARNING_PATTERN,
+ *     /deprecated/i,
+ *     'rate limit'
+ *   ]
+ * })
+ * ```
+ *
+ * @public
+ */
+export const EXPERIMENTAL_API_WARNING_PATTERN = /This is an experimental API version/
```
judofyr: There's no real difference between `EXPERIMENTAL_API_WARNING_PATTERN` and `EXPERIMENTAL_API_WARNING` since they both do substring match. Maybe just remove this one?

src/types.ts
```diff
@@ -111,6 +111,7 @@ export interface ClientConfig {
   headers?: Record<string, string>
 
   ignoreBrowserTokenWarning?: boolean
+  ignoreWarnings?: string | RegExp | Array<string | RegExp>
```
judofyr, line 137: Let's add some documentation here (or somewhere else)? Not sure why we don't have it for the other ones…

Especially the behavior that a `string` does a substring match (and not an exact match) is quite crucial.

src/http/request.ts
```diff
@@ -17,14 +17,39 @@ const httpError = {
   },
 }
 
-function printWarnings() {
+function printWarnings(config: {ignoreWarnings?: string | RegExp | Array<string | RegExp>} = {}) {
   const seen: Record<string, boolean> = {}
+
+  // Helper function to check if a warning should be ignored
+  const shouldIgnoreWarning = (message: string): boolean => {
+    if (!config.ignoreWarnings) return false
```
judofyr: This means that `ignoreWarnings: ''` end up not ignoring anything while `ignoreWarnings: ['']` ignores everything. A bit of an edge case, but would be easily fixed by `config.ignoreWarnings === undefined`)

judofyr (commented):
Implementation itself looks sensible! A few comments.

src/types.ts
kmelve: Yeah, I wasn't sure about this one tbh. I'll remove it.

src/http/request.ts
kmelve: Good catch! Fixed.

src/types.ts
kmelve: Totally!

judofyr (approved):

PR #1108, opened 2025-06-23
feat: version.create action supports optional `baseId` and `versionId` instead of `document`

jordanl17:
As of the latest content lake API, `version.create` server action now supports an optional `baseId` and `versionId` instead of a `document`. eg

```
client.createVersion({
  publishedId: 'my-document',
  baseId: 'versions.summer-sale.my-document',
  releaseId: 'winter-sale'
})
```
will create the document:
```
{
  _id: 'versions.winter-sale.my-document',
  <fields copied from versions.summer-sale.my-document>,
  _system: {
    base: {
      id: 'versions.summer-sale.my-document'
    }
  }
}
```

When `baseId` is provided, then the API will copy from the document at that Id path, creating a new document under the `versionId`.

This is a recommended approach over defining `document` as when using `baseId`, the resultant document will contain a `_system.base.id` property, which will allow for inheritance and head tracing.

### DIFF
--- a/src/SanityClient.ts
+++ b/src/SanityClient.ts
@@ -1477,7 +1477,8 @@ export class SanityClient {
    * Creates a new version of a published document.
    *
    * @remarks
-   * * Requires a document with a `_type` property.
+   * * The preferred approach is to use `baseId` to refer to the existing published document, but it is also possible to provide a complete `document` instead.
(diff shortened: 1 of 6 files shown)


RitaDias (approved):
I have nothing to point in this, if anyone else does feel free to jump in, but otherwise I feel comfortable giving a green

PR #1109, opened 2025-06-25
docs(views): add initial reference docs

danielgroves:
Docs for the views client. I've tried to keep this as light as possible while communicating the functionality so maintainance isn't too bad given this overlaps with the sanityClient in places. Conventions and patterns are copied form elsewhere in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2316,7 +2316,6 @@ const client = createViewClient({
 })
 
 client.fetch('viewId', 'count(*)')
-        .fetch('viewId', 'count(*)')
         .then((data) => console.log(`Number of documents: ${data}`))
         .catch(console.error);
 ```
--- a/README.md
+++ b/README.md
@@ -2380,7 +2380,8 @@ const config: ViewClientConfig = {
   token: process.env.SANITY_API_TOKEN, // this is required to use the override functionality
   viewOverrides: [
     {
-      id: 'vw2x09caysNaZdjksWEsToG43meSg', // the view id you are developing
+      resourceType: ViewResourceType.View,
+      resourceId: 'vw2x09caysNaZdjksWEsToG43meSg', // the view id you are developing
       connections: [
         {
           query: '*[_type == "location"]', // the query to use for this connection - this is what you would deploy later
--- a/README.md
+++ b/README.md
@@ -142,6 +142,15 @@ export async function updateDocumentTitle(_id, title) {
       - [Archive Release Action](#archive-release)
       - [Unarchive Release Action](#unarchive-release)
       - [Delete Release Action](#delete-release)
+  - [Views API](#views-api)
+    - [Quickstart](#quickstart-1)
+    - [Creating a views client instance](#creating-a-views-client-instance)
+      - [ESM](#esm-1)
+      - [CommonJS](#commonjs-1)
+      - [TypeScript](#typescript-1)
+      - [Other frameworks, type safety, etc](#other-frameworks-type-safety-etc)
+    - [Using a view](#using-a-view)
+    - [Developing a view](#developing-a-view)
 - [License](#license)
 - [From `v5`](#from-v5)
   - [The default `useCdn` is changed to `true`](#the-default-usecdn-is-changed-to-true)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -142,6 +142,14 @@ export async function updateDocumentTitle(_id, title) {
       - [Archive Release Action](#archive-release)
       - [Unarchive Release Action](#unarchive-release)
       - [Delete Release Action](#delete-release)
+- [Views API](#views-api)
```
rankers: Should this be nested under API?

README.md
```diff
@@ -2260,6 +2268,152 @@ const result = await client.agent.action.patch({
 })
 ```
 
+## Views API
+
+### Quickstart
+
+```js
+import {createViewClient} from '@sanity/client/views'
```
rankers, line 2277: Should we just show simple case for quickstart against a deployed view and leave overrides for developing a view?

README.md
danielgroves: Sure, can relocate it

README.md
```diff
@@ -2260,6 +2269,139 @@ const result = await client.agent.action.patch({
...
+
+```js
+const { createViewClient } = require('@sanity/client/views')
+
+const client = createViewClient({
+  apiVersion: '2025-06-25',  // use current date (YYYY-MM-DD) to target the latest API version. Note: this should always be hard coded. Setting API version based on a dynamic value (e.g. new Date()) may break your application at a random point in the future.
+})
+
+client.fetch('viewId', 'count(*)')
+        .fetch('viewId', 'count(*)')
```
rankers: Duplicate fetch here

README.md
```diff
@@ -2260,6 +2269,139 @@ const result = await client.agent.action.patch({
...
+
+```ts
+import {createViewClient, ViewClientConfig} from '@sanity/client/views'
+
+const config: ViewClientConfig = {
+  apiVersion: '2025-06-25', // use current date (YYYY-MM-DD) to target the latest API version. Note: this should always be hard coded. Setting API version based on a dynamic value (e.g. new Date()) may break your application at a random point in the future.
+}
+const client = createViewClient(config)
+
+const data = await client.fetch<number>('viewId', 'count(*)')
```
rankers, line 2333: How would this typed fetch work for something thats not a number? For instance just a query to fetch a document? Feels like we're veering into Typescript territory when we don't have full support here, should we drop this example?

README.md
```diff
@@ -2260,6 +2269,139 @@ const result = await client.agent.action.patch({
...
+
+To do this we must specify the new option `viewOverrides` on instantiation of the client. Note that a token is required to access this development workflow. 
+
+```js
+const config: ViewClientConfig = {
+  apiVersion: '2025-06-25', // use current date (YYYY-MM-DD) to target the latest API version. Note: this should always be hard coded. Setting API version based on a dynamic value (e.g. new Date()) may break your application at a random point in the future.
+  token: process.env.SANITY_API_TOKEN, // this is required to use the override functionality
+  viewOverrides: [
+    {
+      id: 'vw2x09caysNaZdjksWEsToG43meSg', // the view id you are developing
```
rankers: Should this be `viewId` similar to elsewhere we've referenced a viewId also should we just make placeholders for `resourceId`?

rankers (changes requested):

README.md
danielgroves: Its take any type here, I jsut used the same example as in the sanityClient. 

Another common way to express something like this in TS docs is `client.fetch<T>`, however happy to drop if you prefer 👍

rankers (approved):

README.md
```diff
@@ -2260,6 +2269,139 @@ const result = await client.agent.action.patch({
...
+
+```ts
+import {createViewClient, ViewClientConfig} from '@sanity/client/views'
+
+const config: ViewClientConfig = {
+  apiVersion: '2025-06-25', // use current date (YYYY-MM-DD) to target the latest API version. Note: this should always be hard coded. Setting API version based on a dynamic value (e.g. new Date()) may break your application at a random point in the future.
+}
+const client = createViewClient(config)
+
+const data = await client.fe\tch<number>('viewId', 'count(*)')
```
rankers: Backslash in fetch?

PR #1112, opened 2025-07-14
feat: add `isHttpError` method (and shared `HttpError` interface)

rexxars:
### Description

Kept re-implementing this in a bunch of different places - when a client request throws and your TS settings are strict, error is always `unknown`. Quite often you're just interested in knowing that the error was actually an HTTP error and what HTTP error code it had. 

This introduces a `isHttpError` utility function to determine this - it does the type narrowing for you. Both the `ClientError` and `ServerError` adhere to the now defined `HttpError` interface.

### What to review

Changes make sense? Sufficient testing?

### Testing

New tests added for the utility function.

### DIFF
--- a/src/defineCreateClient.ts
+++ b/src/defineCreateClient.ts
@@ -21,6 +21,8 @@ export {
   ClientError,
   CorsOriginError,
   formatQueryParseError,
+  type HttpError,
+  isHttpError,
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
makes sense to me

PR #1126, opened 2025-08-12
fix: fixes interface for project API response

binoy14:
Please spot check me if the interface is actually correct

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -451,10 +451,17 @@ export type DatasetsResponse = {
   tags: string[]
 }[]
 
+interface SanityRole {
+  name: string
+  title: string
+  description: string
+}
+
 /** @public */
 export interface SanityProjectMember {
   id: string
   role: string
+  roles: SanityRole[]
   isRobot: boolean
   isCurrentUser: boolean
 }
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -752,6 +752,7 @@ describe('client', async () => {
             role: 'administrator',
             isCurrentUser: true,
             isRobot: false,
+            roles: [{name: 'administrator', title: 'Administrator', description: 'Administrator'}],
           },
         ],
       }
@@ -774,6 +775,7 @@ describe('client', async () => {
             role: 'administrator',
             isCurrentUser: true,
             isRobot: false,
+            roles: [{name: 'administrator', title: 'Administrator', description: 'Administrator'}],
           },
         ],
       }


binoy14:
* **#1126** <a href="https://app.graphite.dev/github/pr/sanity-io/client/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/client/1126?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

rexxars:
The client currently uses the `apiVersion` passed to the configuration for _any_ API request, which means typing the responses is actually kind of hard since we don't currently have a way to defined API-versioned type definitions (made harder by the fact that API versions aren't "hard defined" (v2, v3) but date based.

Unfortunately the `role`/`roles` properties seems to depend on the API version (the type def is actually already wrong):
- `/v1/:project/<id>` returns only a `role` property (not `roles`)
- `/v:date/:project/<id>` returns only `roles`, not `role`

I don't see why we couldn't have v1 return both `role` _and_ `roles` since it's backwards compatible, so one solution would be to update the backend to do so. Another would be to use a static API version for these type of API calls that are generally not meant to be user-controllable. That's a bit of a bigger change in behavior though.

binoy14:
> The client currently uses the `apiVersion` passed to the configuration for _any_ API request, which means typing the responses is actually kind of hard since we don't currently have a way to defined API-versioned type definitions (made harder by the fact that API versions aren't "hard defined" (v2, v3) but date based.
> 
> Unfortunately the `role`/`roles` properties seems to depend on the API version (the type def is actually already wrong):
> 
> * `/v1/:project/<id>` returns only a `role` property (not `roles`)
> * `/v:date/:project/<id>` returns only `roles`, not `role`
> 
> I don't see why we couldn't have v1 return both `role` _and_ `roles` since it's backwards compatible, so one solution would be to update the backend to do so. Another would be to use a static API version for these type of API calls that are generally not meant to be user-controllable. That's a bit of a bigger change in behavior though.

hmm I think the better thing is to update the backend. But this an interesting thing to think thru for any future inconsistencies

binoy14:
Need a better solve, closing for now

PR #1130, opened 2025-08-22
feat: add includeAllVersions option to getDocument method

kmelve:
## Summary
Add support for `includeAllVersions: boolean` parameter to `client.getDocument()` method following the same patterns as other endpoint-specific parameters like `tag`, `signal`, and `releaseId`.

## Changes
- ✅ Added `includeAllVersions?: boolean` parameter to `_getDocument` function options
- ✅ Pass parameter as query string when defined (not undefined)  
- ✅ Updated TypeScript interfaces for both Observable and Promise-based clients
- ✅ Added comprehensive tests covering `true`, `false`, and `undefined` cases
- ✅ Follows existing code patterns for endpoint-specific parameters

## Test plan
- [x] All existing tests pass (902 tests)
- [x] New tests for `includeAllVersions` functionality pass
- [x] Code passes linting with no warnings
- [x] TypeScript compilation succeeds with no errors
- [x] Implementation follows existing code patterns and conventions

The parameter behaves idomatically:
- When `includeAllVersions: true` - adds `?includeAllVersions=true` to API request
- When `includeAllVersions: false` - adds `?includeAllVersions=false` to API request  
- When `undefined` or not provided - no query parameter is added

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/SanityClient.ts
+++ b/src/SanityClient.ts
@@ -262,7 +262,17 @@ export class ObservableSanityClient {
       includeAllVersions?: boolean
     },
   ): Observable<SanityDocument<R> | undefined | SanityDocument<R>[]> {
-    return dataMethods._getDocument<R>(this, this.#httpRequest, id, options as any)
+    // Implementation needs to handle union type safely
+    if (options?.includeAllVersions === true) {
+      return dataMethods._getDocument<R>(this, this.#httpRequest, id, {
+        ...options,
+        includeAllVersions: true,
+      })
+    }
+    return dataMethods._getDocument<R>(this, this.#httpRequest, id, {
+      ...options,
+      includeAllVersions: false,
+    })
   }
 
   /**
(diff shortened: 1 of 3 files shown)


rexxars (changes requested):
- Tests needs to return multiple documents
- Tests needs to ensure that the return value is correct (should be an array of documents instead of a single document)
- `getDocument` method needs overloads when `includeAllVersions: true` is passed that indicates that the return value is an array, not a single document or undefined

stipsan (approved):
Okay ~~Claude~~ _Clanky_ Code it looks like you've addressed @rexxars's concerns 👍

kmelve:
> Okay ~Claude~ _Clanky_ Code it looks like you've addressed @rexxars's concerns 👍

It was under my supervision though!

rexxars (approved):

PR #1134, opened 2025-08-29
feat: add media library video client

sjelfull:
### TL;DR

Added Media Library API support, start with video playback functionality.

### What changed?

- Added a new `mediaLibrary` client interface with a `video` submodule
- Implemented `getPlaybackInfo()` method to retrieve video playback URLs and metadata
- Added typed `getPlaybackTokens()` helper to extract authentication tokens for secured videos
- Added support for various video transformation options (thumbnails, animated previews, etc.)
- Added type definitions for the Media Library API
- Updated README with usage examples and documentation

### How to test?

```
// Basic usage with video asset ID
const playbackInfo = await client.mediaLibrary.video.getPlaybackInfo('video-30rh9U3GDEK3ToiId1Zje4uvalC-mp4')

// With transformations
const playbackInfo = await client.mediaLibrary.video.getPlaybackInfo('video-30rh9U3GDEK3ToiId1Zje4uvalC-mp4', {
  transformations: {
    thumbnail: { width: 300, format: 'webp', fit: 'smartcrop' },
    animated: { width: 200, fps: 15, format: 'webp' }
  },
  expiration: 3600
})

// Using Global Dataset Reference (GDR)
const playbackInfo = await client.mediaLibrary.video.getPlaybackInfo({
  _ref: 'media-library:mlZxz9rvqf76:30rh9U3GDEK3ToiId1Zje4uvalC'
})

// Extract tokens for secured videos
import {getPlaybackTokens} from '@sanity/client/media-library'
const tokens = getPlaybackTokens(playbackInfo)
```

```
// console.log(tokens)
{
   "stream": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9...",
   "thumbnail": "eyJ0a2VuIjoiVGh1bWJuYWlsVG9rZW4tMTIz...",
   "animated": "eyJ0a2VuIjoiQW5pbWF0ZWRUb2tlbi1kZWY...",
   "storyboard": "eyJ0a2VuIjoiU3Rvcnlib2FyZFRva2VuLTc4..."
}
```

### Why make this change?

This change adds programmatic access to the Media Library, allowing developers to retrieve video playback information. The implementation supports both public and signed videos, with options for customizing video transformations like thumbnails and animated previews.

### Questions

- Does "utility" functions like `getPlaybackTokens`, that parses a playback response and gets you a typed object with the tokens you later need to pass to a player like the [Mux React Player](https://www.mux.com/docs/guides/player-advanced-usage#use-signed-urls) makes sense to put here?
- Are we okay with adding a new export `@sanity/client/media-library`?

### DIFF
--- a/src/mediaLibrary/MediaLibraryVideoClient.ts
+++ b/src/mediaLibrary/MediaLibraryVideoClient.ts
@@ -6,11 +6,8 @@ import type {
   HttpRequest,
   MediaLibraryAssetInstanceIdentifier,
   MediaLibraryPlaybackInfoOptions,
+  SanityReference,
   VideoPlaybackInfo,
-  VideoPlaybackInfoItem,
-  VideoPlaybackInfoItemSigned,
-  VideoPlaybackInfoSigned,
-  VideoPlaybackTokens,
 } from '../types'
 
 /** @internal */
@@ -32,8 +29,18 @@ export class ObservableMediaLibraryVideoClient {
     assetIdentifier: MediaLibraryAssetInstanceIdentifier,
     options: MediaLibraryPlaybackInfoOptions = {},
   ): Observable<VideoPlaybackInfo> {
+    const configMediaLibraryId = this.#client.config()['~experimental_resource']?.id
+
     const {instanceId, libraryId} = parseAssetInstanceId(assetIdentifier)
-    const uri = buildVideoPlaybackInfoUrl(instanceId, libraryId)
+    const effectiveLibraryId = libraryId || configMediaLibraryId
+
+    if (!effectiveLibraryId) {
+      throw new Error(
+        'Could not determine Media Library ID - you need to provide a valid Media Library ID in the client config or a Media Library GDR',
+      )
+    }
+
+    const uri = buildVideoPlaybackInfoUrl(instanceId, effectiveLibraryId)
     const queryParams = buildQueryParams(options)
 
     return _request<VideoPlaybackInfo>(this.#client, this.#httpRequest, {
@@ -41,25 +48,6 @@ export class ObservableMediaLibraryVideoClient {
       uri,
       query: queryParams,
     })
-  }
-
-  /**
-   * Extract playback tokens from signed video playback info
-   *
-   * @param playbackInfo - The playback info response
-   * @returns Object containing playback, thumbnail, and storyboard tokens if signed, undefined otherwise
-   */
-  getPlaybackTokens(playbackInfo: VideoPlaybackInfo): VideoPlaybackTokens | undefined {
-    if (isSignedPlaybackInfo(playbackInfo)) {
-      return {
-        playback: playbackInfo.stream.token,
-        thumbnail: playbackInfo.thumbnail.token,
-        storyboard: playbackInfo.storyboard.token,
-        animated: playbackInfo.animated.token,
-      }
-    }
-
-    return undefined
   }
 }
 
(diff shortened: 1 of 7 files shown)


sjelfull:
* **#1134** <a href="https://app.graphite.dev/github/pr/sanity-io/client/1134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/client/1134?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/mediaLibrary/MediaLibraryVideoClient.ts
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * Extract playback tokens from signed video playback info
+   *
+   * @param playbackInfo - The playback info response
+   * @returns Object containing playback, thumbnail, and storyboard tokens if signed, undefined otherwise
+   */
+  getPlaybackTokens(playbackInfo: VideoPlaybackInfo): VideoPlaybackTokens | undefined {
+    if (isSignedPlaybackInfo(playbackInfo)) {
+      return {
+        playback: playbackInfo.stream.token,
```
rdunk: What's the reasoning behind `stream` being aliased to `playback` here? I expected that the shape of this object would align with the properties in `VideoPlaybackInfo`.

src/mediaLibrary/MediaLibraryVideoClient.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  throw new Error('Invalid asset identifier: must be a string or an object with a _ref property')
+}
+
+function buildVideoPlaybackInfoUrl(instanceId: string, libraryId?: string): string {
+  const effectiveLibraryId = libraryId || 'default'
+  return `/media-libraries/${effectiveLibraryId}/video/${instanceId}/playback-info`
+}
```
rdunk: Not sure if I'm missing something here but as I understand it `getPlaybackInfo` calls this function to determine the playback URL, and the `assetIdentifier` that `getPlaybackInfo` accepts needs to be a GDR or `libraryId` will be `undefined` here. In that case the library ID will be `'default'`. Should it not be using the `id` passed via `~experimental_resource`?

src/types.ts
```diff
@@ -1722,3 +1722,146 @@ export type {
...
+  /** Storyboard (contact sheet) options. */
+  storyboard?: StoryboardTransformOptions
+}
+
+/**
+ * Backwards compatibility alias.
+ * @deprecated Use MediaLibraryVideoPlaybackTransformations instead.
+ * @public
+ */
+export type MediaLibraryPlaybackTransformations = MediaLibraryVideoPlaybackTransformations
```
rdunk: What's the backwards compatibility for here? Is there existing usage of this type from somewhere else?

src/mediaLibrary/MediaLibraryVideoClient.ts
```diff
@@ -0,0 +1,217 @@
...
+
+function parseAssetInstanceId(assetIdentifier: MediaLibraryAssetInstanceIdentifier): {
+  instanceId: string
+  libraryId?: string
+} {
+  if (typeof assetIdentifier === 'string') {
+    // Handle video-prefixed asset instance ID
+    if (assetIdentifier.startsWith('video-')) {
+      return {instanceId: assetIdentifier}
+    }
```
rdunk: Is this check redundant if the returned value is the same as the fallback?

rdunk (commented):
This is great overall, just a few questions and comments.

I understand a bit better what we discussed yesterday now. `getPlaybackInfo` makes total sense here, but I see that `getPlaybackTokens` feels maybe a bit out of place, and should maybe rightly be exported as a utility function instead, possibly from a different export path like `@sanity/client/media-library`?

src/mediaLibrary/MediaLibraryVideoClient.ts
sjelfull: This might be me overindexing on using the Mux player, where the key is `playback`. I'll change it to use `stream`. TBH not sure we are naming it `stream` consistently across surfaces.

src/types.ts
sjelfull: Ignore this, leftover from refactoring

src/mediaLibrary/MediaLibraryVideoClient.ts
rdunk: Oh, well I suppose that makes sense if this utility is used primarily for the Mux player?

src/mediaLibrary/MediaLibraryVideoClient.ts
sjelfull: Good point. I think I want to reorder this logic a bit, so that it either checks for a `video-` prefixed instance id string or continues with the other logic.

src/mediaLibrary/MediaLibraryVideoClient.ts
sjelfull: Ah yes, I thought I fixed this. Will update.

sjelfull: I think the nuance here is that you need to make sure it is not using a project hostname, and so we want you to configure it with `~experimental_resource` - but you can theoretically pass in a GDR that points to a different ML id than the one configured for the client.

You are right in that it should never allow it to be undefined in the end.

src/mediaLibrary/MediaLibraryVideoClient.ts
sjelfull: For now, that is true. We currently recommend the Mux player moment because it has all the built-in functionality, but it can be used with any HLS based player. Long term we want to white label the Mux player somehow. Most importantly I think we just want this to be consistent.

rdunk (approved):
LGTM!

PR #1136, opened 2025-09-04
fix: move extra options check after first connect attempt

cngonzalez:
### Description

This PR updates the `client.live.events()` API to do a CORS check _after_ first attempting to connect, rather than before. The initial, explicit `OPTIONS` check in the old code was causing issues because it doesn't fully replicate browser preflight behavior, leading to false negative CORS failures (especially in the SDK).

By just attempting to connect and do the explicit check _after_ failure, we can still get the helpful error message but not run into other issues.

Here's a test application with the failure with this particular change: https://cors-repro-to-delete.sanity.dev/
And here's a new application working as we would like it to: https://www.sanity.io/@oY7HUIM8i/application/dg6kst7me8j16e0cucyact1l/

### What to review
The logic changes aren't huge. They are:
1. Add a catchError handler to run the CORS check, rather than concatting the stream.
2. Also run the CORS check on reconnect events. I found in practice that connection errors are [converted](https://github.com/sanity-io/client/blob/19176aa78643b2ae53de9ee301016e05a099d634/src/data/eventsource.ts#L154) to `reconnect`s and the connection would remain open and receiving `reconnect` events infinitely. Checking these short-circuits this. I'd love confirmation that this isn't going to cause complications somewhere else. 

### Testing
Tests remain largely the same. One test had to be changed because we couldn't intercept the initial `OPTIONS` check, so it is now using nock instead.
Some assertions went down, because there are `expect` blocks in callbacks that were running twice -- once for the initial `OPTIONS` check, and once for the EventSource `GET`.

I added an additional test to ensure we weren't silencing other legitimate errors. I'm not sure how useful it is because of the reconnect behavior.

### DIFF
--- a/src/data/live.ts
+++ b/src/data/live.ts
@@ -1,4 +1,4 @@
-import {catchError, concat, EMPTY, mergeMap, Observable, of} from 'rxjs'
+import {catchError, mergeMap, Observable, of} from 'rxjs'
 import {finalize, map} from 'rxjs/operators'
 
 import {CorsOriginError} from '../http/errors'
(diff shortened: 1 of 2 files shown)


stipsan (approved):
If memory serves the reason the CORS check was done up front rather than after errors were due to not being sure how to best deal with the reconnect events. Looks like you've solved that here 🙌 

Next I hope we can get a `/ping` endpoint or something up and running so we can properly differentiate CORS errors from firewalls and network failure:
image

rexxars (approved):

cngonzalez:
> If memory serves the reason the CORS check was done up front rather than after errors were due to not being sure how to best deal with the reconnect events. Looks like you've solved that here 🙌

Still not the perfect approach, as you say. We'll keep eyes on it to see if it gets noisy (I don't think `reconnect`s are meant to happen too too often?). It also may still flake for the SDK (if we send the explicit OPTIONS check on a reconnect we'll get the old error) but we'll keep eyes on it.

PR #1139, opened 2025-09-12
wip: new resource-based client

judofyr:
This is a proposal for a new API for working with Content Operations.

See https://github.com/sanity-io/client/blob/resource-based-client/docs/resource-guide.md for an explanation.

Here's a list of use cases that we're trying to solve. There's still quite a lot missing.

- [x] Tree-shakable.
- [x] Support for custom methods on resources.
- [x] _Precise_ multi-schema support.
- [x] Make it possible for other packages (e.g. SDK) to provide generic querying support while keeping type safety.
- [ ] Expand `PlainConfig` with additional functionality that we have in the client.
- [ ] Expose mutate and action as a reusable primitive.
- [ ] Expose mutate and action helpers.
- [ ] Typed projections.
- [ ] Typed parameters.
- [ ] `get` multiple documents in one go.
- [ ] `get` to also return if a document was omitted due to access control.

### DIFF
--- a/docs/resource-guide.md
+++ b/docs/resource-guide.md
@@ -1,0 +1,183 @@
+# Resource-based Client
+
+The resource-based client is a new variant of Sanity Client
+
+- Focused on **Content Operations** where you want to modify content within the Sanity platform
+  (as opposed to _Content Distribution_ where you're serving content to your customers).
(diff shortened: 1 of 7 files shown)


judofyr:
No intention of merging this. This was just an experiment.

PR #1141, opened 2025-09-16
fix: remove resource guard from live client

cngonzalez:
FIXES SDK-733

### Description
There's no longer much need to have a resource guard in the live content API, since, as far as I can tell, live events are available for resources (ref https://api.sanity.io/v2025-05-06/media-libraries/mlPGY7BEqt52/live/events and https://api.sanity.io/vX/canvases/cag5gSK37IGV/live/events)

### What to review
Any issues in removing this?

### Testing
Don't believe this is required.

### DIFF
--- a/src/data/live.ts
+++ b/src/data/live.ts
@@ -13,7 +13,6 @@ import type {
   SyncTag,
 } from '../types'
 import {shareReplayLatest} from '../util/shareReplayLatest'
-import * as validate from '../validators'
 import {_getDataUrl} from './dataMethods'
 import {connectEventSource} from './eventsource'
 import {eventSourcePolyfill} from './eventsourcePolyfill'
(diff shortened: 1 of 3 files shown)


ryanbonial:
@cngonzalez did you check if live events for global resources work on Document resources too?

cngonzalez:
> @cngonzalez did you check if live events for global resources work on Document resources too?

looks like we're still 404ing, yeah https://api.sanity.io/vX/projects/hprmeffw/datasets/production/live/events. Unless I have the format wrong. Apologies, thought these were all the same!

sgulseth (changes requested):
Can we add a test that it creates a working and expected request?

cngonzalez:
> Can we add a test that it creates a working and expected request?

got it, sorry about that!

sgulseth (approved):

PR #1143, opened 2025-09-29
feat: add X-Sanity-Lineage header

snocorp:
### Description

As part of a larger change to detect recursion and prevent infinite loops in functions, we will attach a new header to outgoing requests that will indicate the recursion depth of the function execution.

### What to review

Ensure this works server side and does not impact browser usage.

### Testing

Verified by pointing the library at a local server to verify the header is attached when the env var is set. Used a small vite project to verify this works as expected in the browser (i.e. no header)


src/http/request.ts
```diff
@@ -104,3 +110,18 @@ function shouldRetry(err: any, attempt: number, options: any) {
...
+ * @returns headers middleware that will add the lineage header to requests
+ */
+function lineage(defaultLineage?: string) {
+  const lineageHeaders: Record<string, unknown> = {}
+
+  if (typeof globalThis !== 'undefined' && globalThis.process?.env?.X_SANITY_LINEAGE) {
+    lineageHeaders['X-Sanity-Lineage'] = globalThis.process.env.X_SANITY_LINEAGE
+  } else if (defaultLineage) {
+    lineageHeaders['X-Sanity-Lineage'] = defaultLineage
+  }
```
sgulseth: This doesn't make much sense in a browser environment. I would rather include this as part of the envMiddleware here:
https://github.com/sanity-io/client/blob/e9dabf773b54264e5f1bee1b6d8bdd4d2d88daeb/src/http/nodeMiddleware.ts

sgulseth (dismissed):
I think we should move the this into the specific runtime environment instead of setting it for all runtime envs.
IWBN with some tests as well - we have a few that asserts headers etc

src/http/request.ts
snocorp: You are correct but unfortunately it really difficult to get the config into this middleware from that file. It pretty much has to be static without a fairly large refactoring. The reason I need the config is because when the AI agent runs these command on the user's behalf, they create a client that should also use the lineage from the incoming http request.

rexxars: Maybe I am overlooking something, but have a look at https://github.com/sanity-io/client/pull/1144 and see if that makes sense?

rexxars (approved):

src/http/request.ts
sgulseth: nice!

PR #1151, opened 2025-11-04
fix: make releaseId optional in createVersion method

jordanl17:
### Description
`releaseId` can validly be `undefined` for the `createVersion` method. In this case, as the existing comment notes, the version created will be a draft.

Wrongly the type definition did not mark `releaseId` as optional. This PR resolves this.

### Testing
There was already a test covering this, but silly-ly I had used `as any` cast and so this issue didn't surface. Corrected the test to verify the type fix.


RitaDias (dismissed):
Nice!

PR #1154, opened 2025-11-06
feat: add support for resuming listener events

bjoerge:
Note: This enableResume is currently only available on vX so this is not ready to be merged yet, but otherwise it's ready for review.

### Description
Adds `enableResume: boolean` to `ListenerOptions`. This instructs the listen-endpoint to push down events based on the `Last-Event-Id` header, and adds two more listener events:

- `welcomeback` – when the listener has reconnected successfully. This event is emitted immediately when listener has successfully reconnected, and any resumed mutation events will arrive after.
- `reset` –  signals that listener can't resume events. Consumers should respond to this event by resetting local state.

### What to review
- ~What apiVersion should we require for this?~ (this will be available on all apiVersions with `enableResume=true`)

### Testing

Unit tests added.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1182,7 +1182,7 @@ export type ResetEvent = {
 }
 
 /** @public */
-export type ListenEvent<R extends Record<string, Any>> =
+export type ListenEvent<R extends Record<string, Any> = Record<string, Any>> =
   | MutationEvent<R>
   | ReconnectEvent
   | WelcomeBackEvent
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -1139,12 +1140,33 @@ export type WelcomeEvent = {
...
+ *
+ * @public
+ */
+export type WelcomeBackEvent = {
+  type: 'welcomeback'
+  listenerName: string
+}
+
+/**
+ * The listener has reconnected and successfully resumed from where it left off
```
mwain: Comment is wrong, should be `The listener has reconnected and failed to resume` 
or something similar

bjoerge: Ah, ty, is it always after a _reconnect_? I was under the impression that (at least in theory) it could be emitted at any point

mwain: We always send a welcome/welcomeback message first. Yes the reset message could be emitted at any point in the stream (after welcomes/welcomeback)

sgulseth (dismissed):

mwain (approved):

mwain (dismissed):
LGTM

sgulseth (approved):

PR #1157, opened 2025-11-12
feat: use expression operations

pedrobonamin:
### Description
Adds support for variants expressions instead of `anyOf`


pedrobonamin:
We don't need this, closing

PR #1159, opened 2025-11-17
feat(listen): return event type based on options.events

bjoerge:
### Description
Currently, `client.listen()` returns Observable<ListenEvent>, where `ListenEvent` is a union all events that listen _can potentially_ emit:
https://github.com/sanity-io/client/blob/7e956b84c93d75a0a07d501598e222087dcfa6d0/src/types.ts#L1143-L1149

However, if you pass e.g. `{events: ['mutation', 'welcome']}`, in reality you'll only get `Observable<MutationEvent | WelcomeEvent>`

This PR narrows the returned event type based on the list of events names provided by the listener options.

Also adds `open` to the list of possible events to subscribe to. Don't think this event is super useful for most cases (better off using `welcome` which represents e2e connection with the backend stream), but could be of interest for debugging, etc.

Also removes `ChannelErrorEvent` and `DisconnectEvent` from the `ListenEvent` union since these are thrown as errors on the observable, not emitted as events.

### Testing
This mainly affects the type system, runtime behavior should not change.

See added type tests: [`fix/map-listen-events-to-event-types?expand=1`#diff-b88c5f3110](https://github.com/sanity-io/client/compare/fix/map-listen-events-to-event-types?expand=1#diff-b88c5f3110425a87b8eecabf1167328edc676c9c7f74f8018a84edbb4fc8b2fb)

### DIFF
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -4,6 +4,7 @@ import {filter, map} from 'rxjs/operators'
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
 import {
   type Any,
+  type ListenEvent,
   type ListenEventName,
   type ListenOptions,
   type ListenParams,
@@ -78,7 +79,8 @@ export type ListenEventFromOptions<
 > = Opts extends ListenOptions
   ? Opts['events'] extends ListenEventName[]
     ? MapListenEventNamesToListenEvents<R, Opts['events']>
-    : MutationEvent<R>
+    : // fall back to ListenEvent if opts events is present, but we can't infer the literal event names
+      ListenEvent<R>
   : MutationEvent<R>
 
 /**
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -4,10 +4,13 @@ import {filter, map} from 'rxjs/operators'
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
 import {
   type Any,
-  type ListenEvent,
+  type ListenEventName,
   type ListenOptions,
   type ListenParams,
   type MutationEvent,
+  type OpenEvent,
+  type ReconnectEvent,
+  type WelcomeEvent,
 } from '../types'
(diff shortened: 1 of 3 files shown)


src/data/listen.ts
```diff
@@ -36,6 +41,52 @@ const defaultOptions = {
...
+    : E extends 'mutation'
+      ? MutationEvent<R>
+      : E extends 'reconnect'
+        ? ReconnectEvent
+        : E extends 'open'
+          ? OpenEvent
+          : E extends 'disconnect'
+            ? DisconnectEvent
+            : E extends 'channelError'
+              ? ChannelErrorEvent
```
bjoerge: gonna remove these now that I discovered they are not emitted, hang on

sgulseth (dismissed):
Discussed async that this might be a breaking change from a type perspective, but we don't expect many usages of the removed message type out there.

src/data/listen.ts
```diff
@@ -36,6 +40,49 @@ const defaultOptions = {
...
+ *
+ * @public
+ */
+export type ListenEventFromOptions<
+  R extends Record<string, Any> = Record<string, Any>,
+  Opts extends ListenOptions | undefined = undefined,
+> = Opts extends ListenOptions
+  ? Opts['events'] extends ListenEventName[]
+    ? MapListenEventNamesToListenEvents<R, Opts['events']>
+    : // fall back to ListenEvent if opts events is present, but we can't infer the literal event names
```
bjoerge, line 82: cc @sgulseth – think this is our best option for now (see also the added [type test](https://github.com/sanity-io/client/pull/1159/files#diff-b88c5f3110425a87b8eecabf1167328edc676c9c7f74f8018a84edbb4fc8b2fbR25-R31) covering this).

sgulseth: sgtm! At least we are giving the same interface as today

sgulseth (approved):

PR #1161, opened 2025-11-18
feat(listen)!: enable resume by default

bjoerge:

### DIFF
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -12,8 +12,6 @@ import {
   type OpenEvent,
   type ReconnectEvent,
   type ResetEvent,
-  type ResumableListenEventNames,
-  type ResumableListenOptions,
   type WelcomeBackEvent,
   type WelcomeEvent,
 } from '../types'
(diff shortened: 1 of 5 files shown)


bjoerge:
Closing this as I think it's better to leave resumability off by default. I'd gladly have my mind changed, but my current thinking is that whether you want resumability or not depends heavily on use case:
- Consumers that act on realtime events would trigger stale actions. Notifications, counters would be fired retroactively in a burst on reconnect, which is rarely what you want.
- Resume is an optimization. The catch-up window is bounded to ~5–10 minutes (IIRC), so a resume can always fail and emit reset. Any consumer that cares about gaps therefore need to implement a refetch-on-reset path regardless.
- Even if documented as a breaking change, the change is subtle and timing bound, and might not surfaced as a compile time/TS error after upgrading. In resume mode a reconnect emits `welcomeback` or `reset` rather than a second `welcome`, so consumers refetching on a repeat welcome silently lose that branch.

PR #1178, opened 2026-01-05
chore(media-library): add thumbhash support to AssetMetadataType

danielgroves:
### Description

Dependency of adding support to main sanity package. New property being added to the media library asset instance. 

### Testing

BEGIN_COMMIT_OVERRIDE
feat(media-library): add thumbhash support to AssetMetadataType
fix(types): thumbhash -> thumbHash
END_COMMIT_OVERRIDE

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -370,7 +370,7 @@ export interface SanityImageAssetDocument extends SanityAssetDocument {
     isOpaque: boolean
     lqip?: string
     blurHash?: string
-    thumbhash?: string
+    thumbHash?: string
     dimensions: {
       _type: 'sanity.imageDimensions'
       aspectRatio: number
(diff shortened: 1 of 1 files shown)


sjelfull (changes requested):
Should add it to `SanityImageAssetDocument` type as well?

src/types.ts
```diff
@@ -369,6 +370,7 @@ export interface SanityImageAssetDocument extends SanityAssetDocument {
     isOpaque: boolean
     lqip?: string
     blurHash?: string
+    thumbhash?: string
```
sjelfull:
```suggestion
    thumbHash?: string
```

sjelfull (changes requested):

sjelfull (approved):

PR #1184, opened 2026-02-14
feat: improve get-it HTTP stacktraces

jonahsnider:
Related to AIGRO-4516
Blocked until sanity-io/get-it#556 is merged

### Description

Please see sanity-io/get-it#556 for the original context on the issue

get-it gives very vague stacktraces which don't include the original source of the error.
`@sanity/client` does some `Observable` stuff which makes it hard to resolve that purely from get-it.
This PR captures the state of the call stack and then passes that down to get-it, so that if an error is thrown it includes the full context.

Here's an example of an error stacktrace with and without this change:

```diff
diff --git a/before.txt b/after.txt
index 1fe090c..56d5a46 100644
--- a/before.txt
+++ b/after.txt
@@ -1,11 +1,20 @@
 ClientError: Not Found - Project with ID "foo" not found
     at onResponse (/app/mellon/node_modules/client/dist/index.js:182:13)
     at applyMiddleware (/app/mellon/node_modules/get-it/src/util/middlewareReducer.ts:9:40)
-    at onResponse (/app/mellon/node_modules/get-it/src/createRequester.ts:62:22)
-    at <anonymous> (/app/mellon/node_modules/get-it/src/createRequester.ts:104:55)
-    at callback (/app/mellon/node_modules/get-it/src/request/node-request.ts:145:81)
-    at cb (/app/mellon/node_modules/get-it/src/request/node-request.ts:246:14)
+    at onResponse (/app/mellon/node_modules/get-it/src/createRequester.ts:68:22)
+    at <anonymous> (/app/mellon/node_modules/get-it/src/createRequester.ts:121:55)
+    at callback (/app/mellon/node_modules/get-it/src/request/node-request.ts:147:81)
+    at cb (/app/mellon/node_modules/get-it/src/request/node-request.ts:250:14)
     at Transform.<anonymous> (/app/mellon/node_modules/get-it/src/request/node/simpleConcat.ts:8:13)
     at Object.onceWrapper (node:events:622:28)
     at Transform.emit (node:events:508:28)
     at endReadableNT (/app/mellon/node_modules/get-it/node_modules/through2/node_modules/readable-stream/lib/_stream_readable.js:1003:12)
+    at _requestObservable (/app/mellon/node_modules/client/dist/index.js:913:25)
+    at _request (/app/mellon/node_modules/client/dist/index.js:938:10)
+    at DatasetsClient.list (/app/mellon/node_modules/client/dist/index.js:1442:7)
+    at _tool (/app/mellon/src/mcp/tools/datasets/updateDatasetTool.ts:38:42)
+    at <anonymous> (/app/mellon/src/mcp/utils/response.ts:160:20)
+    at McpServer.executeToolHandler (/app/mellon/node_modules/.pnpm/@modelcontextprotocol+sdk@1.24.3_zod@3.25.76/node_modules/@modelcontextprotocol/sdk/src/server/mcp.ts:351:42)
+    at <anonymous> (/app/mellon/node_modules/.pnpm/@modelcontextprotocol+sdk@1.24.3_zod@3.25.76/node_modules/@modelcontextprotocol/sdk/src/server/mcp.ts:213:43)
+    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
+    at async wrappedHandler (/app/mellon/node_modules/.pnpm/@modelcontextprotocol+sdk@1.24.3_zod@3.25.76/node_modules/@modelcontextprotocol/sdk/src/server/index.ts:262:32)
```

Without this change, there's no indication as to where the error is originating from.
But when we capture the call stack before making the request, we can see that `updateDatasetTool.ts` is the file that actually initiates the HTTP request that's failing!

### Testing

Tested locally with `pnpm link`


rexxars (dismissed):
Looks good! Would be cool to see a test that proves that a stack trace from get-it now is more readable, but not a blocker

jonahsnider:
> Looks good! Would be cool to see a test that proves that a stack trace from get-it now is more readable, but not a blocker

Test added. Confirmed it fails on main and passes on this branch.

rexxars (dismissed):

rexxars (approved):

PR #1186, opened 2026-02-17
feat(types): add VideoRenditionInfo types with typed resolution

sjelfull:
## Summary

Adds rendition type definitions to `@sanity/client` to match the backend response shape.

Part of DAM-1038

### Types (`src/types.ts`)
- `VideoRenditionInfoPublic` — `url` + `resolution`
- `VideoRenditionInfoSigned` — extends public with `token` + `expiresAt`
- `VideoRenditionInfo` — union type
- `renditions?` field added to `VideoPlaybackInfo`

The `resolution` property is typed as `'1080p' | '480p' | '270p' | (string & {})` — gives IDE autocomplete for known values while accepting any string for future resolutions.

### Exports (`src/media-library.ts`)
Re-exports all 3 rendition types from the media library entry point.

### Documentation (`README.md`)
- Added rendition examples to both public and signed playback info responses
- Added "Downloading MP4 renditions" section with usage example

## Note
Subtitle types have been split into a separate PR (#1191) for independent rollout.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1978,7 +1978,31 @@ export interface VideoPlaybackInfoItemSigned extends VideoPlaybackInfoItemPublic
 export type VideoPlaybackInfoItem = VideoPlaybackInfoItemPublic | VideoPlaybackInfoItemSigned
 
 /** @public */
-export interface VideoPlaybackInfo<T extends VideoPlaybackInfoItem = VideoPlaybackInfoItem> {
+export interface VideoRenditionInfoPublic {
+  /** URL to the MP4 rendition (redirects to CDN) */
+  url: string
+  /** Resolution identifier, e.g. "1080p", "480p", "270p" */
+  resolution: '1080p' | '480p' | '270p' | (string & {})
+}
+
+/** @public */
+export interface VideoRenditionInfoSigned extends VideoRenditionInfoPublic {
+  /** Authentication token for signed playback */
+  token: string
+  /** Token expiration time in ISO 8601 format */
+  expiresAt: string
+}
+
+/** @public */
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -1986,6 +2005,7 @@ export interface VideoPlaybackInfo<T extends VideoPlaybackInfoItem = VideoPlayba
   stream: T
   duration: number
   aspectRatio: number
+  renditions?: VideoRenditionInfo[]
 }
```
rdunk: Just one suggestion.

In the current changes `renditions` is always typed as `VideoRenditionInfo[]` regardless of whether the playback info is signed or public.

We can add a second generic param derived from T which means that when you have a `VideoPlaybackInfoSigned` type the `renditions` property automatically narrows to `VideoRenditionInfoSigned[]`, otherwise you need manual checking of e.g. `token` to narrow it.

```suggestion
export interface VideoPlaybackInfo<
  T extends VideoPlaybackInfoItem = VideoPlaybackInfoItem,
  R extends VideoRenditionInfo = T extends VideoPlaybackInfoItemSigned
    ? VideoRenditionInfoSigned
    : VideoRenditionInfo,
> {
  id: string
  thumbnail: T
  animated: T
  storyboard: T
  stream: T
  duration: number
  aspectRatio: number
  renditions?: R[]
}
```

rdunk (changes requested):
One suggestion for an improvement, it's not crucial but slightly better DX I think.

src/types.ts
sjelfull: Great suggestion — applied! The second generic param now narrows renditions automatically when the playback info type is signed.

rdunk (dismissed):
Great! LGTM

sjelfull:
@rdunk I rebase, seems like I have to re-request review.

PR #1190, opened 2026-02-26
fix(stega): add mux video machine fields to default filter denylist

kmelve:
## Summary

- Add 13 machine-generated fields to the stega `filterDefault` denylist to prevent stega encoding from breaking rendering and logic in Mux video integrations and image placeholders
- Fields added: `aspect_ratio`, `aspectRatio`, `blurHash`, `thumbHash`, `encoding_tier`, `master_access`, `max_resolution_tier`, `max_stored_resolution`, `mp4_support`, `resolution_tier`, `upload_id`, `video_quality`, `video_resolution`
- None of these are user-editable content, so filtering them has zero impact on visual editing

## Context

The Mux video plugin (`sanity-plugin-mux-input`) stores machine-generated fields from the Mux API verbatim in `mux.videoAsset` documents. When stega-encoded, these fields break downstream consumers — most critically `aspect_ratio`/`aspectRatio` which corrupts CSS `aspect-ratio` calculations, and various enum/config fields used in conditional logic.

Similarly, `blurHash` and `thumbHash` are machine-generated image placeholder hashes (analogous to the already-denylisted `lqip`) that break when stega characters are injected.

## Test plan

- [x] All 146 filterDefault tests pass (13 new test cases added)
- [x] Full stega test suite passes (180 tests)


kmelve:
I'm not sure if this is the best approach to be honest. But Mux is the main way to use Video through Sanity and not escaping these fields made problems for our preview. I wish there was a way to declare "machine fields" from the backend through CSM.

PR #1191, opened 2026-02-26
feat(types): add VideoSubtitleInfo types to playback info

sjelfull:
## Summary

Adds subtitle type definitions for the video playback info response, split out from #1186 (renditions PR) to keep concerns separate.

### Types added
- `VideoSubtitleInfoPublic` — trackId, languageCode, url
- `VideoSubtitleInfoSigned` — extends public with token + expiresAt
- `VideoSubtitleInfo` — union type
- `subtitles?: VideoSubtitleInfo[]` field on `VideoPlaybackInfo`

Re-exported from `@sanity/client/media-library`.

### DIFF
--- a/src/media-library.ts
+++ b/src/media-library.ts
@@ -7,6 +7,7 @@ import type {
 } from './types'
 
 export type {VideoRenditionInfo, VideoRenditionInfoPublic, VideoRenditionInfoSigned} from './types'
+export type {VideoSubtitleInfo, VideoSubtitleInfoPublic, VideoSubtitleInfoSigned} from './types'
 
 /**
  * Check if a playback info item (stream/thumbnail/etc) has a signed token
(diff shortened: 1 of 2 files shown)


src/media-library.ts
```diff
@@ -7,6 +7,7 @@ import type {
 } from './types'
 
 export type {VideoRenditionInfo, VideoRenditionInfoPublic, VideoRenditionInfoSigned} from './types'
+export type {VideoSubtitleInfo, VideoSubtitleInfoPublic, VideoSubtitleInfoSigned} from './types'
```
rdunk, line 10: nit: could mash these exports together

rdunk (approved):
LGTM. Only question is if this is user facing/needs a README update. But fine to merge!

PR #1200, opened 2026-03-09
feat(projects): add `onlyExplicitMembership` option to `projects.list()`

binoy14:
## Summary
- Adds `onlyExplicitMembership` boolean option to `projects.list()` on both `ObservableProjectsClient` and `ProjectsClient`
- When set to `true`, filters projects to only those where the user has explicit membership (as opposed to implicit membership via organization)
- Parameter is omitted from the API request when `false` or unset (default behavior unchanged)

## Test plan
- [x] Added test: `onlyExplicitMembership: true` sends `?onlyExplicitMembership=true`
- [x] Added test: `onlyExplicitMembership: false` does not include the parameter
- [x] Added test: combines correctly with other options (`organizationId`)
- [x] All 488 existing tests pass
- [x] No type errors

### DIFF
--- a/src/projects/ProjectsClient.ts
+++ b/src/projects/ProjectsClient.ts
@@ -19,15 +19,22 @@ export class ObservableProjectsClient {
    * @param options - Options for the list request
    *   - `includeMembers` - Whether to include members in the response (default: true)
    *   - `organizationId` - ID of the organization to fetch projects for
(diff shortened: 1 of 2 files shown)


rexxars (approved):
Theoretically the defaults for the parameter can change between API versions (it doesnt currently), but we can fix this later if need be.

PR #1149, opened 2025-10-22
fix(perf): improving stega encoding

cocoviper:
Recreated the stega library in Sanity's codebase to apply improvements on the code. If anyone has a contact at Vercel I'd be more than happy to make a pull request for their codebase as well.

Stega Encoding was also slow on really big queries (100+ mappings) . The resolveMappings function is improved. Performance is 2x faster than latest on benchmark test . And it is up to 4x faster on bigger mappings (tested on Rona inc. results not shared in this commit)

### Description

What changes are introduced? Improved Stega encoding (10x faster) , Improved resolveMappings Function reducing unnecessary loops and array duplication through filter + sort to find the longest matching key.
Why are these changes introduced? Performance
What issue(s) does this solve? Crashing on NextJS when a query with 1000+ mappings is processed

### What to review

What steps should the reviewer take in order to review? Test Stega on presentation tool and with Vercel (the new algorithm is compatible with the old encoding)
What parts/flows of the application/packages/tooling is affected? Stega Encoding

### Testing

Did you add sufficient testing for this change? Tests were already in place for benchmarking and stega encoding

PR #1204, opened 2026-03-19
fix(stega): allow zero-width space characters in content

### LINKED ISSUE #1202: Stega decoding fails when original content ends with one or more zero-width space characters
**Describe the bug**

If content pasted into the Sanity studio contains trailing characters such as `U+200B ZERO WIDTH SPACE`, that field will fail to be editable in the Presentation Tool and an error will be shown in the console such as:

```
Failed to decode stega for string: (...) with the original error:  Error: Encoded data has invalid length
```

or

```
Failed to decode stega for string: (...) with the original error:  SyntaxError: Unexpected token '(...)', "(...)"... is not valid JSON
```

We've experienced this problem organically with content that our authors pasted into the studio from other sources.

**To Reproduce**

1. Paste the text between the quotes here - "foo​" - into a field in the Sanity Studio. (Note: there is a single zero-width space character after the letter o. The string could also be represented in javascript as `'foo.\u200b'`.)
2. Open the document in the Presentation Tool
3. Notice that the field is not outlined for editing
4. Notice the `Failed to decode stega` error in the console.

Alternatively, the following test can reproduce the error inside this repo:

```javascript
import {expect, test} from 'vitest'

import {stegaCombine,stegaDecode} from '../../src/stega/stega'

test('it handles trailing zero-width space characters', () => {
  const payload = 'foo.\u200b'
  const editInfo = JSON.stringify({origin: 'sanity.io', href: '/studio'})
  const encoded = stegaCombine(payload, editInfo)
  expect(encoded).not.toEqual(payload)
  const decoded = stegaDecode(encoded)
  expect(decoded.origin).toEqual('sanity.io')
  expect(decoded.href).toEqual('/studio')
})
```

**Expected behavior**

The original text should survive stega encoding and decoding without errors, and visual editing should work as normal.

**Which versions of Sanity are you using?**

```
@sanity/cli (global)               6.1.5 (latest: 6.1.7)
@sanity/cross-dataset-duplicator   1.5.1 (up to date)
@sanity/icons                      3.7.4 (up to date)
@sanity/orderable-document-list    1.4.2 (latest: 1.5.1)
@sanity/ui                        3.1.13 (latest: 3.1.14)
@sanity/vision                    5.16.0 (up to date)
sanity                            5.16.0 (up to date)
@sanity/eslint-config-studio       5.0.2 (latest: 6.0.0)
```

**What operating system are you using?**

macOS 26.3.1

**Which versions of Node.js / npm are you running?**

```
11.3.0
v24.1.0
```

stipsan:
### Description

Fixes #1202, vercel fixed it on their end in `@vercel/stega@1.1.0`.
In `@vercel/stega@1.0.0` they also backported the perf work from @cocoviper in #1149, allowing us to remove our inlined stega code and rely on the utils from `@vercel/stega` once more.

### What to review

Looks good?
Ran `npm run benchmark` to confirm that perf is equivalent after removing the inlined vercel stega logic.

### Testing

If the tests pass we're good.


bjoerge (approved):
LGTM

PR #1208, opened 2026-03-31
feat: support intercepting requests via internal requestHandler config

bjoerge:
### Description

  Adds a `requestHandler` config option that lets consumers intercept HTTP requests for logging, custom headers, auth token refresh, metrics, rate limiting, etc. My primary use case for this is to improve handling of expired token, cors checks etc. in the Studio. Happy to make it hidden/private if we want to reduce public exposure/dependence of request options, etc.

I did consider the deprecated `requester` config, but that seems to tie us to get-it internals, which I'd like to avoid.

Example:
```ts
const client = createClient({
  projectId: 'my-project',
  dataset: 'production',
  requestHandler: (request, defaultRequester, client) => {
    return defaultRequester({
      ...request,
      headers: {...request.headers, 'CustomHeader': 'value'},
    })
  },
})
```

  The handler receives resolved request options, a `defaultRequester` that executes the request through the normal get-it pipeline, and a `client` instance without a `requestHandler` for making  side requests (e.g. token refresh on 401) without recursive interception. Consumers can modify requests, transform responses, or skip the default requester entirely.

- Works with both promise and observable APIs
- `withConfig()` can override or remove the handler (replaces, does not wrap)
- Does not apply to `client.listen()`
- The bare `client` is lazily created on first handler invocation

### What to review 
- `src/types.ts` — `RequestHandler` type and `requestHandler` config option (does the naming make sense?)
- `src/SanityClient.ts` — handler applied in constructors of both client classes, `#originalHttpRequest` stored so `clone()`/`withConfig()` can re-wrap
 
### Testing
New test file covering: custom header injection, no-op passthrough, response transformation, short-circuit, `withConfig()` override/remove/preserve, bare client for side requests, and observable API compatibility.

Note: implemented by claude, reviewed by me.

### DIFF
--- a/src/SanityClient.ts
+++ b/src/SanityClient.ts
@@ -43,6 +43,7 @@ import type {
   RawQuerylessQueryResponse,
   RawQueryResponse,
   RawRequestOptions,
+  RequestOptions,
   SanityDocument,
   SanityDocumentStub,
   SingleActionResult,
@@ -87,6 +88,7 @@ export class ObservableSanityClient {
    * Private properties
    */
   #clientConfig: InitializedClientConfig
+  #originalHttpRequest: HttpRequest
   #httpRequest: HttpRequest
 
   /**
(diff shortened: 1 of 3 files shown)


rexxars (changes requested):
Can we mark this as internal? We've got some upcoming work to make the client more tree shakeable and compose differently, and this would be yet another breaking change once we do. I don't mind the studio using it, as we control that - but would rather not have it be used externally.

bjoerge:
> Can we mark this as internal?

Absolutely! Excited about the upcoming work 🚀 

Update: marked as internal in a5eb90f

rexxars (approved):

PR #1209, opened 2026-04-01
feat(live): add waitFor option to defer events until Sanity Function processing

adoprog:
### Description

Adds a new `waitFor` option to LiveClient.events() that accepts 'function' as its value. When set, the live events stream delays delivery until after a Sanity Function has processed the event and called the callback endpoint. 

When `waitFor` is omitted (default), events continue to be delivered immediately — no behavior change for existing consumers.


stipsan (approved):
🥳 💖

PR #1212, opened 2026-04-14
feat(error): expose trace id from backend responses

kbrabrand:
## Summary

Expose the W3C `traceparent` trace ID on `ClientError` and `ServerError` for correlating client errors with backend traces.

## What changed

`extractErrorProps` parses the `traceparent` header and extracts the trace ID segment. Added `traceId` to `ErrorProps` and both error classes.

## Why

Makes it easy to go from a client-side error to the corresponding backend trace without digging through response headers.

Kong returns a `traceparent` containing the information we need but to be able to surface it in a sensible way we need to expose it through the error from the client. The end goal is to make it easier to debug end-user issues by making it easier for them to find and provide us with the trace ID.

## Testing

Unit tests for trace ID extraction on both error classes, including missing/absent header cases.

### DIFF
--- a/src/http/errors.ts
+++ b/src/http/errors.ts
@@ -95,14 +95,14 @@ function extractErrorProps(res: Any, context?: HttpContext): ErrorProps {
     response: res,
     statusCode: res.statusCode,
     responseBody: stringifyBody(body, res),
-    traceId: traceId(res),
+    traceId: extractTraceId(res),
     message: '',
     details: undefined as Any,
   }
 
   // Fall back early if we didn't get a JSON object returned as expected
   if (!isRecord(body)) {
-    props.message = httpErrorMessage(res, body)
+    props.message = `${httpErrorMessage(res, body)}${formatTraceId(props.traceId)}`
     return props
   }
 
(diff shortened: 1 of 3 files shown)


sgulseth (commented):
talked sync: let's expose traceId in the formatted/printed error messages as well

kbrabrand:
> talked sync: let's expose traceId in the formatted/printed error messages as well

Exposed now for the different error cases. Positive: will most likely surface the traceId most everywhere, downside: putting it in the message makes it a little harder to make the error pretty downstream, but given that these are error cases it's probably a fair trade-off and worth it to make the trace id visible in most places.

Tests added.

sgulseth (approved):
Looks good to me! @rexxars any opinions on error here?

src/http/errors.ts
```diff
@@ -196,17 +199,22 @@ export function isQueryParseError(error: object): error is QueryParseError {
...
  * @public
  */
-export function formatQueryParseError(error: QueryParseError, tag?: string | null) {
+export function formatQueryParseError(
+  error: QueryParseError,
+  tag?: string | null,
+  traceId?: string,
+) {
   const {query, start, end, description} = error
+  const withTraceId = traceId ? `\nTraceId: ${traceId}` : ''
```
rexxars, line 208: Appreciate aligning with tag here, but feel it will be helpful if all traceId mentions use the same formatting (minor)

```suggestion
  const withTraceId = traceId ? `\n(traceId: ${traceId}`) : ''
```

rexxars (dismissed):
Looks good - personally I'd have the formatting be the same shape for GROQ errors, but very minor. Remember that if this is to work in browsers, the header needs to be added to `Access-Control-Expose-Headers`, otherwise it won't show up.

sgulseth (approved):

PR #1215, opened 2026-04-27
feat: add documentsExists for batch existence checks

kbrabrand:
## Summary

Adds `documentsExists(ids, options?)` to `SanityClient` and `ObservableSanityClient`, returning a `Set<string>` of IDs that exist.

## What changed

Hits `/data/doc/<ids>?excludeContent=true` and derives existence from the response's `omitted` array — only `reason: "existence"` excludes an ID; other reasons (e.g. `permissions`) are treated as existing. IDs are percent-encoded and batched at 100 per request. Options: `signal`, `tag`.

## Why

A dedicated method avoids branching `getDocument`'s return type on a flag and gives callers a clear API for existence/permission pre-checks without fetching content.

## Testing

Unit tests cover all-present, mixed omissions by reason, single/empty input, missing `omitted` field, percent-encoding, tag forwarding, HTTP errors, and >100-id batching.


src/SanityClient.ts
```diff
@@ -324,6 +324,21 @@ export class ObservableSanityClient {
     return dataMethods._getDocuments<R>(this, this.#httpRequest, ids, options)
   }
 
+  /**
+   * Checks which of the given document IDs exist. Returns a set of the IDs that exist —
+   * IDs the API reports as omitted with reason `existence` are excluded from the set;
+   * IDs omitted for any other reason remain in the set.
```
sgulseth: This level of details is a bit verbose here? It should explain why to use this method, not what it's doing under the hood?
How about something like:
```suggestion
   * Convenient and bandwidth efficient method of checking wether a set of document IDs exists.
   * Returns a set of the IDs that exist.
```
?

sgulseth (approved):

PR #1217, opened 2026-05-21
feat!: require node 22.12 or higher, use fetch

rexxars:
### Description

Moves `@sanity/client` onto [get-it v9](https://github.com/sanity-io/get-it/pull/576), which rewrites the HTTP transport on native `fetch`. Raises the Node floor to 22.12 and makes the package ESM-only. This is a major release.

Worth stating up front: this is deliberately not the ideal end-state of a refactored client. It is a derisked transport and dependency upgrade that keeps the public surface as stable as possible while swapping the foundation underneath. Follow-up work is already scoped for modernizing the client itself: tree-shakeable parts, better typing, and smaller bundle size, plus a full browser test suite along the lines of what get-it now has (testing in real browsers rather than relying on happy-dom alone).

#### Why

The transport stack this replaces dates back to when making an HTTP request meant picking through each environment's quirks (the rationale in sanity-io/get-it#576 applies here wholesale, since end users only ever experience get-it through this client):

- **Three transports, three sets of bugs.** `XMLHttpRequest` in browsers, `node:http` on the server, and an opt-in fetch path. Every feature was built and tested three times, and the paths drifted apart in subtle ways. There is now one transport: `fetch`, stable in Node, Bun, Deno, workers, and every current browser.
- **The dependencies were old.** The stack pulled in packages, several unmaintained, that papered over gaps the platform has since closed: redirect following, proxy tunneling, XHR shims, and an SSE implementation that talked `node:http` directly. The transport now leans on the platform; `undici` is used in Node only, for proxy support.
- **Middleware cost more than it returned.** Errors from the server used to arrive buried under a dozen frames of middleware channel subscriptions. Stack traces now point at your code.
- **The platform caught up.** Cancellation is plain `AbortController`/`AbortSignal` end to end, timeouts build on `AbortSignal.timeout()`, and streams are Web streams.

#### What changed

Transport (`src/http/`, `src/data/`):

- Requests flow through get-it v9's promise-based `createRequester`. The v8 multi-event observable pipeline is gone: data methods are promise-first with `_*Observable` siblings, and the observable API is a thin wrapper that aborts the underlying fetch on unsubscribe.
- Error semantics are preserved: 4xx/5xx still throw `ClientError`/`ServerError` with the same shape. One visible difference: errors now carry the real `statusText`, so messages read `HTTP 400 Bad Request` rather than `HTTP 400`.
- Retry semantics are preserved: GET/HEAD and `/data/query` requests retry 429/502/503 with backoff, and `maxRetries: 0` still opts out.
- The public request options (`withCredentials`, `timeout: 0` to disable, `maxRedirects`, object-form `fetch`, ...) are translated at a single boundary: `requestOptions` projects them plus the client config straight into get-it v9 fetch options, and everything below that speaks v9. There is no legacy v8 options shape left in the pipeline. This also surfaced and fixed a hidden bug: the `lineage` config option was silently dropped by the old double translation, so its header only worked via the env var (and its test never asserted the header).

Next.js App Router compatibility:

- The `cache` and `next` options on `client.fetch()` (and the deprecated object-form `fetch` client config) are forwarded into the actual fetch init. get-it v9's `fetch` option only accepts a function, so the init extras travel through request `meta` and are merged into whichever fetch implementation is effective (`applyFetchInit` in `src/http/request.ts`). This was silently dropped in an earlier iteration of this branch; it is now regression-tested.
- Signals and Request Memoization: Next.js' patched fetch opts a request out of React Request Memoization whenever `init.signal` is present ([dedupe-fetch.ts](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/dedupe-fetch.ts), verified current through the Next 16 canary - this is intentional, since a memoized response cannot safely be tied to one caller's abort lifetime). get-it v9 implements timeouts via `AbortSignal.timeout()`, including a 120s default when no timeout is set, so naively every query would carry a signal and silently disable memoization - the v8 fetch transport only attached a signal when `useAbortSignal` was truthy. The adapter honors that flag again: queries without a caller-provided signal reach fetch signal-free (`timeout: false` on the transport), and the timeout is enforced as a rejection-only race (`withSoftTimeout`) that leaves the underlying request running, same as v8. Caller-provided signals flow through untouched and opt out of memoization, which is Next's documented contract.

Server-sent events (`listen()` / `live.events()`):

- `@sanity/eventsource` is replaced with upstream [`eventsource@4`](https://github.com/EventSource/eventsource), which is fetch-based, so SSE now flows through the same transport as everything else.
- Unlike before, the polyfill is no longer conditionally loaded: it is always used, in every environment. It is well maintained (by yours truly), builds on the same modern platform primitives as get-it, and weighs about 2 kB min+gzip. Always using it removes a whole family of code-path variants, and means any configuration of the fetch layer (proxying, custom headers, a custom fetch implementation) flows the exact same way for SSE as for regular requests. The native browser `EventSource` cannot do any of that (it does not even support setting an `Authorization` header).
- The `proxy` config option now reaches SSE connections too, and `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` keep working in Node via undici.
- Connections rejected with a permanent 4xx (for example an expired token) stop reconnecting and error with `ConnectionFailedError` (check `.status`); 408, 429, 5xx and network drops keep reconnecting. This carries #1226 over to the new stack.

Asset uploads:

- Upload progress events are gone from the general transport, for two reasons. Most use cases never consumed them, and threading progress through every request added complexity to every code path for the benefit of one. And `fetch` has no widely supported way to observe upload progress in the first place (request-body streaming is not usable cross-browser).
- The exception is browsers, where a dedicated XHR-based path (`src/http/browserUpload.ts`, lazily imported) still emits progress events for `assets.upload()`. This exists because the studio is powered by this client, and for large uploads you still want a progress bar. We consider it an unfortunate hack and want to remove it as soon as a fetch-based alternative is viable.
- Node and edge runtimes emit only the terminal response event.

Packaging:

- ESM-only: all CJS entry points are gone. The Node 22.12 floor makes this a soft break, `require('@sanity/client')` keeps working via `require(esm)`, and a CJS smoke test guards that.
- The exports map is inverted to mirror get-it's convention: the platform-neutral fetch build is the default (`dist/index.js`) and the undici/Node build is the special case (`dist/index.node.js`). Explicit `workerd`/`worker`/`deno`/`bun`/`react-server`/`sanity-function` conditions are listed ahead of `node` so runtimes that co-activate the `node` condition still get the fetch build. A packaging-resolution guard in `test/exports.test.ts` models Node's condition resolution and asserts workers can never resolve the undici build.
- `undici` is kept out of the browser and UMD bundles entirely. The final UMD bundle is ~33 kB gzipped, slightly below the v8 baseline (35.1 kB), with the ESM builds essentially unchanged.

#### Breaking changes

- Node >= 22.12 is required (Node 20 left LTS in April 2026). The specific floor follows the studio: the studio minimum usually sets the precedent for the rest of the Sanity ecosystem, the CLI uses the same bar, and get-it and now the client follow it. In hindsight 22.19 would have been the better pick when the studio release went out (it would have allowed the latest undici, for instance), but that ship has sailed and it is not a major blocker.
- The package is ESM-only. On supported Node versions existing `require()` calls keep working natively.
- The `requester` config option is removed. It existed to swap get-it v8's `Requester` via `.clone()/.use()`, which v9 deleted. Migrate to the `headers` config and/or a custom `fetch`.
- The internal, deprecated `_requestHandler` config hook is removed.
- The `HttpRequestEvent`, `ResponseEvent` and `ProgressEvent` types are removed. Observable `assets.upload()` now emits `UploadEvent<T>` instead; the `type` discriminants are unchanged, but the response event drops the transport-level fields (`statusCode`, `statusMessage`, `headers`, `url`, `method`).
- The `unstable__adapter` / `unstable__environment` re-exports are removed (get-it v9 does not expose them).
- The per-request `proxy` option is removed. Proxying is a client-level concern now: pass `proxy:` on the client config (replaceable via `client.config({proxy})` / `withConfig({proxy})`), or use `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` env vars. This drops a whole layer of per-request indirection; callers that need different proxies can derive a client per proxy with `withConfig`.
- The v8-style `uri` request option is removed; requests are addressed with `url` only, and `url` is now required. It takes the same API-relative path `uri` did (an absolute URL for the raw `requester` export), so migrating is a rename: `client.request({url: '/ping'})`. The `uri` property is gone from the `RawRequestOptions` and `RequestObservableOptions` types, so TypeScript callers catch this at compile time; plain-JS callers passing `uri` (or nothing) get a TypeError stating that request options must include a url.
- Node no longer emits upload progress events; browsers still do, via the XHR path described above.

Full migration notes live in [docs/MIGRATE-v7.md](https://github.com/sanity-io/client/blob/feat/get-it-9/docs/MIGRATE-v7.md) (the README's migration section is now one bullet per version, linking to per-version files under docs/).

### What to review

This branch reads well commit by commit: each commit message documents its rationale and test fallout, and the commits build on each other in order (transport decoupling, get-it v9 swap, eventsource swap, packaging inversion, test-infra migration). Reviewing by area also works:

- `src/http/` for the transport and error mapping (`requestOptions` as the single public-options-to-fetch boundary, `executeRequest`, `httpResponseFromFetch`)
- `src/data/` for the promise-first data methods and the eventsource/fetch injection (`resolveEventSourceFetch`)
- `package.json` exports plus `test/exports.test.ts` for the packaging story
- `test/helpers/` for the mock-fetch infrastructure

Spots that deserve extra scrutiny:

- the exports map key ordering (runtime conditions ahead of `node`)
- `resolveEventSourceFetch`'s base-fetch resolution order (test hook, then the environment's fetch via `resolveFetch(proxy?)`, then global fetch) - SSE must use the same fetch implementation as regular requests
- the retry opt-out now living at `options.meta.maxRetries === 0`
- the signal handling for Next.js Request Memoization (`useAbortSignal` in `requestOptions` plus `withSoftTimeout`): signal-less queries reject on timeout without aborting the underlying request, trading hard cancellation for memoization compatibility

### Testing

- The suite migrated from nock+msw to get-it's built-in `createMockFetch` (nock patches `http.request`, which undici's fetch never touches). A fresh mock is installed per test and injected into clients through the public `resolveFetch` config option - the supported way to swap the transport - so there is no test-only hook in production code.
- Suites, all green: Node (~1000 tests plus type tests), simulated browser (happy-dom 20, which also un-hung the browser suite), vercel-edge, workerd (new; runs the real fetch build in miniflare without `nodejs_compat`, which would catch the undici build leaking in), Deno, Bun, and the Node `require(esm)` interop smoke test.
- As noted above, a real-browser test suite (matching get-it's setup) is planned as follow-up work; happy-dom is a simulation and this branch already hit one of its spec deviations.
- Test studio running the preview release: https://test-studio-git-chore-upgrade-client.sanity.dev/test (branch: https://github.com/sanity-io/sanity/tree/chore/upgrade-client)
- Sanity Function usage: verified working by @filmaj 

---

> [!NOTE]
> **High Risk**
> Major semver bump with Node floor, ESM-only publishing, and CI/runtime matrix changes that gate whether edge, workerd, and Deno consumers still pass.
> 
> **Overview**
> This PR lands the **v8 breaking release** packaging and contributor surface around the fetch-based client: **`engines.node` is `>=22.12.0`**, the package is **ESM-only** (no CJS/`umd` artifacts in CI artifacts or docs), and migration content moves to **`docs/MIGRATE-v4.md`**, **`docs/MIGRATE-v5.md`**, and **`docs/MIGRATE-v7.md`** with the README trimmed (no CommonJS/UMD install examples).
> 
> **CI** is retuned for that story: the build job drops **benchmark** and **`umd/`** from uploaded artifacts; **`ls-engines`** runs as **`--mode virtual --no-dev`** so dev-only engine constraints (e.g. vitest) do not fail published-consumer checks; the cross-OS matrix pins **Node 22.12.0** as the minimum floor instead of `lts/-1`; the standalone **`esm-compatibility`** job is removed; **`edge-runtime`** splits into **`vercel-edge`** and **`workerd-runtime`** with explicit **`dist/`** artifact restore paths; Deno CI upgrades to **setup-deno v2** / **v2.x**; and the **`deno/import_map.json` maintenance workflow** is deleted (Prettier no longer restores that file). **`.gitignore`** swaps **`umd`** for **`deno.lock`**, and **eslint** drops **`umd`** ignores and tightens unused-var rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4578283cf6f32eb24ad1bff615b3cf7d28059fb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


test/stega/stegaEncodeSourceMap.bench.ts
stipsan, line 1: Was deleting this intentional?

package.config.ts
```diff
@@ -3,6 +3,8 @@ import {defineConfig} from '@sanity/pkg-utils'
 export default defineConfig({
   tsconfig: 'tsconfig.dist.json',
 
+  rollup: {hashChunkFileNames: true},
```
stipsan, line 6: `@sanity/pkg-utils` v12 hashes chunk filenames btw

package.json
```diff
@@ -27,58 +27,40 @@
...
       "source": "./src/media-library.ts",
       "import": "./dist/media-library.js",
-      "require": "./dist/media-library.cjs",
       "default": "./dist/media-library.js"
     },
     "./package.json": "./package.json"
   },
-  "main": "./dist/index.cjs",
+  "main": "./dist/index.js",
   "unpkg": "umd/sanityClient.min.js",
```
stipsan: We can probably drop this field, and the umd build itself, while at it

stipsan (dismissed):
Nice work! 💖

test/stega/stegaEncodeSourceMap.bench.ts
rexxars: Yes, Claude and I had a chat about it and decided that this bench is likely no longer relevant? It was added two years ago when the stega encoding started handling portable text differently. Could bring it back if we expect to be changing this implementation - the `@sanity/client-latest` aliased install just adds a bit of noise.

package.json
rexxars: Will do.

test/stega/stegaEncodeSourceMap.bench.ts
stipsan: No worries, it was handy when doing perf work related to how `@vercel/stega` decodes payloads, easy enough to add back if/when needed :)

stipsan (dismissed):

stipsan (approved):

src/http/nodeMiddleware.ts
```diff
@@ -1,41 +1,105 @@
...
+
+function isNodeReadableStream(value: unknown): value is Readable {
+  if (typeof value !== 'object' || value === null) return false
+  if (!('pipe' in value)) return false
+  return typeof value.pipe === 'function'
+}
+
+// One undici-backed fetch per (proxy URL) so callers don't pay the cost of
+// rebuilding a dispatcher on every request when they use the same proxy.
+const proxyFetchCache = new Map<string, ReturnType<typeof createNodeFetch>>()
```
bjoerge, line 21: Do we ever remove entries from this cache? Unclear to me how big this typically/worst case can grow.

package-lock.json
bjoerge, line 1: unrelated to this pr, but: thoughts on moving to pnpm?

bjoerge (commented):
Great stuff!

PR #1226, opened 2026-07-09
fix: stop reconnecting live/listen connections rejected with a 4xx

binoy14:
When an SSE connection (`live.events()` or `listen()`) is rejected with a 4xx — most commonly a 401 from an expired or revoked token — the client reconnects every second, forever. The server keeps rejecting, so the loop never resolves. Request logs show individual apps stuck in this state generating ~86K requests per day each.

- `ConnectionFailedError` now carries the HTTP `status` of the rejected connection when the EventSource implementation exposes it (the Node/headers polyfills do; native `EventSource` does not).
- `onError` errors the stream whenever the error event carries a status, regardless of `readyState` — the polyfills disagree on whether the connection closes before or after the error event is dispatched. This also fixes a pre-existing zombie: statuses the Node polyfill dispatches before closing (eg 501, proxy 52x) previously left a silently dead connection that neither errored nor reconnected.
- `reconnectOnConnectionFailure()` rethrows 4xx failures to the consumer instead of retrying, with 408 (request timeout) and 429 (rate limited) carved out as transient — consistent with the HTTP layer, which retries 429 with backoff. 5xx and status-less failures reconnect as before.

Behavior change: consumers of `live.events()` and `listen()` that previously saw silent `reconnect` events on a 4xx now get an error with `status` set — check it to react (eg refresh auth and resubscribe). Known limitation: native `EventSource` connections (no custom headers, eg cookie auth) expose no status, so 4xx rejections there keep the old reconnect behavior; a backoff/cap for status-less failures is a candidate follow-up.

FIXES SDK-1891

### DIFF
--- a/src/data/eventsource.ts
+++ b/src/data/eventsource.ts
@@ -5,11 +5,25 @@ import {type Any} from '../types'
 
 /**
  * @public
- * Thrown if the EventSource connection could not be established.
- * Note that ConnectionFailedErrors are rare, and disconnects will normally be handled by the EventSource instance itself and emitted as `reconnect` events.
+ * Thrown when the EventSource connection could not be established, or was rejected by the server.
+ * Transient failures (network drops, 5xx, 408, 429) are reconnected internally and emitted as
+ * `reconnect` events; a permanent rejection (any other 4xx, eg an expired token) errors the
+ * stream with this class so consumers can react — check `status` for the rejection code.
  */
 export class ConnectionFailedError extends Error {
   readonly name = 'ConnectionFailedError'
+  /**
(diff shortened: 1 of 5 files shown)


binoy14:
This change is part of the following stack:

- #1226 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

cngonzalez (approved):
not an expert but looks correct to me -- I don't see an issue with adding `status`, and the error handling matches other errors. It does change behavior so I do wonder if this is a `feat` or if we consider the old behavior fully a bug that consumers were handling anyway.

bjoerge (approved):
Makes sense to me, thank you!

PR #1232, opened 2026-07-17
feat: variant querying

juice49:
### Description

This branch adds a `variant` option. It must either be a string representing a variant id, or a set of variant conditions:

```ts
{
  variant: 'someVariantId',
},
```

```ts
{
  variant: {
    market: 'eu',
    currency: 'gbp',
    audience: 'musicians',
  },
},
```

In this initial implementation, only the first variant condition is passed to the request (`market: 'eu'`). We will discuss how the backend handles array values with Content Lake and follow up with a fix. For now, a warning is logged showing which conditions were dropped.

### What to review

New `variant` option and related types.

### Testing

Added unit tests.

### DIFF
--- a/src/data/dataMethods.ts
+++ b/src/data/dataMethods.ts
@@ -549,6 +549,7 @@ export function _dataRequest(
     tag,
     returnQuery,
     perspective: options.perspective,
+    variant: options.variant,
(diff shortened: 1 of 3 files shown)


stipsan (approved):
LGTM! ❤

PR #1234, opened 2026-07-20
feat(stega): add branded string types for stega encoded results

stipsan:
## Problem

[sanity#9303](https://github.com/sanity-io/sanity/issues/9303): with stega enabled, strings contain invisible characters, so `if (imageLocation === 'left')` silently fails while the TypeScript types (`'left' | 'right'`) claim it's safe. The types lie.

## Solution

New opt-in type utilities under `@sanity/client/stega`, all tagged `@beta`, that brand every possibly-stega-encoded string in a query result as "poisoned". Comparing a branded string against a literal is a compile error (TS2367 "no overlap") until it's passed through `stegaClean()`, which strips the brand and recovers the original type. Rendering keeps working: branded strings stay assignable to `string`, so JSX text nodes, template literals and string methods are unaffected (pinned by type tests against real `@types/react`).

```ts
import {type ClientReturnStega, stegaClean} from '@sanity/client/stega'

const post = await client.fetch<ClientReturnStega<typeof query>>(query)

post.imageLocation === 'left' // ❌ compile error
stegaClean(post.imageLocation) === 'left' // ✅ 'left' | 'right'
const node = <h1>{post.title}</h1> // ✅ rendering is fine
```

### New API

- `StegaString<T>`: a branded string, `` `${T} (may contain hidden stega characters)` & {readonly ' stegaBrand': T} ``. The template-literal suffix is what makes literal comparisons, `switch` cases and literal-union assignments error (a plain `string & {brand}` intersection doesn't error on `===`), and it makes the compiler error message self-explanatory. The brand property preserves the original type so `stegaClean` can recover it; it's string-keyed rather than a `unique symbol` so brands stay structurally compatible across duplicated copies of `@sanity/client` in `node_modules`.
- `StegaBranded<T>`: deeply brands strings in a result type, mirroring what the runtime encoder guarantees it never touches:
  - underscore-prefixed keys (`_id`, `_type`, `_createdAt`, `_updatedAt`, `_key`, `_ref`, ...) stay plain, which also preserves discriminated-union narrowing on `_type`
  - `slug.current` patterns and string-valued `slug` keys stay plain
  - Portable Text internals mirror `walkMap`: only span `text` is branded, `style`/`listItem`/`marks`/`markDefs` stay plain
  - everything else is assumed poisoned, even when the runtime `filter` happens to skip it (URLs, dates, `...Id` keys) — a redundant `stegaClean` is harmless, a missing one is a bug
- `ClientReturnStega<GroqString, Fallback>`: drop-in first generic for `client.fetch`, `StegaBranded<ClientReturn<...>>`. Works with today's `fetch` signatures with no changes (an explicit first generic makes `G` default to `string`, so `ClientReturn` resolves to the fallback). Typegen users swap one generic; upstream libraries like `next-sanity` can build a `defineLive({stega: false})` style toggle on top of these exports.
- `stegaBrand(result)`: identity function to re-type already-fetched data where a generic can't be injected (loaders, props boundaries).
- `stegaClean()` now returns `StegaCleaned<Result>`, the deep inverse of `StegaBranded`. For never-branded inputs it stays mutually assignable with the input type (pinned by type tests), so this is non-breaking for existing `@public` usage.

## Testing

- `test/stega/stegaBrand.test-d.ts`: comparison/`switch`/literal-union enforcement, safe keys, `slug.current`, `_type` narrowing, nullable unions, arrays, Portable Text shapes, the `SanityQueries`/typegen fetch flow (incl. `filterResponse: false`), `stegaClean` round-trip and identity, and a large page-builder-shaped fixture.
- `test/stega/react.test-d.tsx`: `<div>{branded}</div>` type-checks against real `@types/react` (new devDependency, `"jsx": "react-jsx"` added to tsconfig), literal-union props correctly error.
- `test/stega/stegaBrand.test.ts`: runtime identity and encode → clean round-trip.
- `npm test` (1050 tests, typecheck clean), `npm run lint`, `npm run prepublishOnly` (strict api-extractor build), `npm run test:next`, `test:browser`, `test:edge-runtime` all pass.

<div><a href="https://cursor.com/agents/bc-314f1106-8316-49f7-bacb-87ec73075514"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-314f1106-8316-49f7-bacb-87ec73075514"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


rexxars (approved):
Very clever!

PR #1235, opened 2026-07-22
feat: add collaboration comments client (`@alpha`)

hermanwikner:
### Description

This PR adds alpha `client.collaboration.comments` methods.

Every Comments API request needs three query params: `organizationId`, `resourceType`, and `resourceId`. A resource is whatever is being commented on. Today that means Canvas (`resourceType=canvas&resourceId=<canvas-id>`) or a dataset (`resourceType=dataset&resourceId=<projectId>.<dataset>`).

The comment methods read those from client config instead of requiring them on every call:

- `collaboration.organizationId` – optional field on `ClientConfig`
- Either an explicit `resource` (`type` / `id`, sent as `resourceType` / `resourceId`), **or** `projectId` + `dataset` (derived as a dataset resource)

Once set, every `collaboration.comments` call adds those query params. Missing `collaboration.organizationId`, or missing both `resource` and `projectId`/`dataset`, throws.

Comments use the same request URL as the rest of the client:

- Explicit `resource` → global API host
- `projectId` / `dataset` without `resource` → project API host (Studio-style clients)

| Method | HTTP |
| --- | --- |
| `create` | `POST /collaboration/comments` |
| `update` | `PATCH /collaboration/comments/:id` |
| `delete` | `DELETE /collaboration/comments/:id` |
| `addReaction` | `POST /collaboration/comments/:id/reactions` |
| `removeReaction` | `DELETE /collaboration/comments/:id/reactions/:shortName` |
| `fetch` | `GET` / `POST /collaboration/comments/query` |
| `listen` | `GET /collaboration/comments/listen` (SSE) |

> [!NOTE]
> Marked `@alpha` while Studio adoption and comment migration are incomplete. These methods use the Comments API, which stores comments in the organization store. Canvas already uses this store, but Studio currently reads and writes comments in add-on datasets. As a result, comments created through this API for `dataset` resources will not appear in Studio until Studio adopts the Comments API and existing comments are migrated.

### Configuring the client

With an explicit resource, comments calls target that resource on the global API host:

```ts

const client = createClient({
  token: '<token>',
  apiVersion: '2026-07-18',
  collaboration: {organizationId: '<org-id>'},
  resource: {type: 'canvas', id: '<canvas-id>'},
})
```

A project-based client only needs `collaboration.organizationId` added. The dataset resource is derived from `projectId` / `dataset`, and requests go to the project API host:

```ts
const client = createClient({
  projectId: '<project-id>',
  dataset: 'production',
  token: '<token>',
  apiVersion: '2026-07-18',
  useCdn: false,
  collaboration: {organizationId: '<org-id>'},
})
```

To target a different resource without rebuilding the client, override `resource` with `withConfig`. Org stays put unless you override it too:

```ts
const datasetClient = client.withConfig({
  resource: {type: 'dataset', id: '<projectId>.<dataset>'},
})

await datasetClient.collaboration.comments.create({/* … */})
```

#### `create`

One method for top-level comments and replies. Top-level needs `target`. A reply needs `parentCommentId` instead (not both). The server copies target/thread from the parent on replies.

```ts
// Top-level comment
await client.collaboration.comments.create({
  message: [/* portable text */],
  target: {documentId: '<document-id>', documentType: 'article'},
})

// Reply
await client.collaboration.comments.create({
  message: [/* portable text */],
  parentCommentId: '<parent-comment-id>',
})
```

Inline comments additionally pass `path` and `range`. The API resolves `range` into `target.path.selection` and `contentSnapshot`; `range` itself is not returned.

#### `update`

Pass `message`, `status` (`open` or `resolved`), and/or `range`. Changing status cascades to replies. A `range` re-anchors an inline comment within the field it already targets; pass `null` to clear the selection and leave a field-level comment.

```ts
await client.collaboration.comments.update('<comment-id>', {
  message: [/* portable text */],
  status: 'resolved',
})

await client.collaboration.comments.update('<comment-id>', {
  range: {start: {_key: 'block-1', offset: 0}, end: {_key: 'block-1', offset: 12}},
})

await client.collaboration.comments.update('<comment-id>', {range: null})
```

#### `addReaction` / `removeReaction`

Comment id first, emoji short name second (for example `:+1:`).

```ts
await client.collaboration.comments.addReaction('<comment-id>', ':+1:')
await client.collaboration.comments.removeReaction('<comment-id>', ':+1:')
```

#### `delete`

Deletes the comment and its replies.

```ts
await client.collaboration.comments.delete('<comment-id>')
```

#### `fetch`

Same shape as `client.fetch(query, params?)`, but against the comments endpoint for the configured resource. Oversized queries are retried as `POST`. Comment documents are `_type == "sanity.comment"`.

Use `getTargetDocumentRef(documentId)` when you need a global document ref (`resourceType:resourceId:publishedId`) for filters. Draft/version ids are normalized to the published id.

```ts
const ref = client.collaboration.comments.getTargetDocumentRef('<document-id>')

await client.collaboration.comments.fetch(
  `*[_type == "sanity.comment" && target.document._ref == $ref && status == "open"] | order(_createdAt desc)[0...50]`,
  {ref},
)
```

#### `listen`

Same call style as `client.listen(query, params?, options?)`, against the comments listen endpoint.

```ts
client.collaboration.comments
  .listen(
    `*[_type == "sanity.comment" && target.document._ref == $ref]`,
    {ref: client.collaboration.comments.getTargetDocumentRef('<document-id>')},
  )
  .subscribe((event) => {
    // …
  })
```

### What to review

- Deriving a dataset resource from `projectId` / `dataset` (no explicit `resource`) so Studio-style clients can use comments on the project API host. Explicit `resource` still wins and stays on the global host.
- `range` on `update` (including `null` to clear). Server recalculates selection/snapshot; range is not returned in responses.
- `getTargetDocumentRef()` normalizes draft/version ids to published. OK as implicit behavior?

---

> [!NOTE]
> **Medium Risk**
> New alpha public API and config semantics (`organizationId`, resource derivation) affect how org-scoped clients behave; listen/query refactors are shared with existing `client.listen`/`fetch` paths.
> 
> **Overview**
> Adds an **alpha** `client.collaboration.comments` (and `client.observable.collaboration.comments`) surface for organization-scoped document comments on Canvas or dataset resources.
> 
> **Client wiring:** `SanityClient` / `ObservableSanityClient` expose `collaboration.comments` with promise and Observable clients. `ClientConfig` adds `collaboration.organizationId`. Calls require `collaboration.organizationId` plus either `resource` or derived `projectId`/`dataset`; every request sends `organizationId`, `resourceType`, and `resourceId`.
> 
> **Capabilities:** CRUD-style writes (`create`, `update`, `delete`), emoji reactions, GROQ `fetch` (GET/POST like `client.fetch`), SSE `listen` (reuses extracted `_connectListenEventSource`), and `getTargetDocumentRef()` for global document refs with draft/version normalization.
> 
> **Internals:** New `src/collaboration/` module with typed comment documents, targets, inline `range`, and mutation response handling (e.g. picking the right document when status cascades to replies). `getQuerySizeLimit`, `MAX_URL_LENGTH`, and listen defaults/options are exported for reuse.
> 
> **Docs & release:** README Collaboration Comments section, changeset minor bump, and broad unit/type tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b375675d5a78c6b30a0e40a61db345eef71e1201. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


snorrees (approved):

snorrees (dismissed):
Initial take is that you are missing actually updates to the REAME.md, and personally I would like to see meaty TSDocs on all public methods simliar to what Agent actions has

src/collaboration/comments.ts
```diff
@@ -0,0 +1,259 @@
...
+  if (!resource) {
+    throw new Error('`resource` must be configured to use collaboration comments')
+  }
+
+  return {
+    organizationId,
+    resourceId: resource.id,
+    resourceType: resource.type,
+  }
+}
```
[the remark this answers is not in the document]
hermanwikner: This is already validated in `validate.resourceConfig()` when the client is initialized. That also runs for `withConfig()`, so an invalid dataset resource ID fails before `resourceQuery()` is called. Adding the same check here would duplicate the validation.

src/collaboration/comments.ts
```diff
@@ -0,0 +1,259 @@
...
+  type CollaborationCommentsStructuredFetchOptions,
+  type CollaborationCommentsStructuredListenOptions,
+  type CollaborationCommentsWriteOptions,
+  type CollaborationCommentUpdate,
+  possibleRequestOptions,
+} from './types'
+
+type Client = SanityClient | ObservableSanityClient
+
+function commentUrl(id: string): string {
```
pepahlavacek, line 37: we should probably check against empty strings here so we don't end up with `delete('')` producing `DELETE /collaboration/comments/`

src/util/getPublishedId.ts
```diff
@@ -0,0 +1,16 @@
+const DRAFTS_PREFIX = 'drafts.'
```
pepahlavacek: this file seems like it duplicates constants from https://github.com/sanity-io/client/pull/1235/changes#diff-7e47bc4c68b55decd1eb5c9b86820ab4d49b34e70bb8c92128b649b7fd43f0abR12-R20

test/collaborationCommentsClient.test.ts
```diff
@@ -0,0 +1,351 @@
...
+
+import {type ClientConfig, type CollaborationCommentDocument, createClient} from '@sanity/client'
+import {firstValueFrom} from 'rxjs'
+import {describe, expect, test} from 'vitest'
+
+import {createSseServer} from './helpers/sseServer'
+
+const apiHost = 'https://api.sanity.url'
+const organizationId = 'org-123'
+const resource = {type: 'canvas' as const, id: 'canvas-123'}
```
pepahlavacek, line 14: should we use something similar to https://github.com/sanity-io/client/blob/main/test/client.test.ts#L315-L335 to test datasets as well and not only canvas?

src/collaboration/comments.ts
```diff
@@ -0,0 +1,259 @@
+import {type Observable, throwError} from 'rxjs'
```
pepahlavacek, line 2: not sure i spotted a test for creating a reply with `parentCommentId`

pepahlavacek (changes requested):
Quick code review before I jump into the things you actually asked to be reviewed 😅 

_filtered down through a human from an AI run_

src/collaboration/comments.ts
hermanwikner: Added one

test/collaborationCommentsClient.test.ts
hermanwikner: Pushed an update

src/collaboration/comments.ts
hermanwikner: Fixed

src/util/getPublishedId.ts
hermanwikner: We can't share these across entry points because that breaks the build. Duplicating the two constants avoids that. Let me know if you can think of a cleaner approach, otherwise I think it's fine to leave this as-is for now!

src/types.ts
```diff
@@ -126,6 +126,12 @@ type ClientConfigResource =
 export interface ClientConfig {
   projectId?: string
   dataset?: string
+  /**
+   * ID of the organization that organization-scoped APIs address: required by
+   * `client.collaboration.comments`, and the owner of the blueprints stack
+   * `functions.invoke()` resolves function names against
+   */
```
pepahlavacek: let's shorten this. there seems to be some unnecessary context

test/client/data.mutations.test.ts
```diff
@@ -367,6 +367,50 @@ describe('mutations', () => {
...
+    ]
+    getActiveMock()
+      .scope(projectHost())
+      .on('POST', '/v1/data/mutate/foo?returnIds=true&visibility=sync', {body: expectedBody})
+      .respond({status: 200, body: {transactionId: 'abc123', results}})
+
+    const res = await getClient().delete({query}, {returnFirst: false, returnDocuments: false})
+    expect(res.documentIds, 'returns all document ids').toEqual(['abc123', 'abc456'])
+    expect(res.transactionId, 'returns transaction ID').toEqual('abc123')
+    expect(res.results, 'returns the raw results').toEqual(results)
```
pepahlavacek: check the whole object

test/collaborationComments.test-d.ts
```diff
@@ -0,0 +1,269 @@
...
+  test('getTargetDocumentRef returns a global document reference', () => {
+    expectTypeOf(comments.getTargetDocumentRef('doc-1')).toEqualTypeOf<
+      CollaborationCommentDocument['target']['document']['_ref']
+    >()
+    expectTypeOf(observableComments.getTargetDocumentRef('doc-1')).toEqualTypeOf<
+      CollaborationCommentDocument['target']['document']['_ref']
+    >()
+  })
+
+  test('writes are not void', () => {
```
pepahlavacek: do we need to still test it? i think just checking the expected shapes should be enough. this probably stayed behind from the rewrite

test/collaborationCommentsClient.test.ts
```diff
@@ -0,0 +1,1161 @@
...
+      .on('DELETE', '/v2026-07-18/collaboration/comments/comment-1/reactions/%3Aheart%3A', {query})
+      .respond({
+        status: 200,
+        body: mutationResponse([{id: 'comment-1', operation: 'update', document: commentDocument}]),
+      })
+
+    await client.collaboration.comments.update('comment-1', {status: 'resolved'}, options)
+    await client.collaboration.comments.delete('comment-1', options)
+    await client.collaboration.comments.addReaction('comment-1', ':heart:', options)
+    await client.collaboration.comments.removeReaction('comment-1', ':heart:', options)
```
pepahlavacek: is this actually checking anything?

src/collaboration/CollaborationCommentsClient.ts
```diff
@@ -0,0 +1,385 @@
...
+   * Fetch comments on the configured resource.
+   *
+   * Takes the same `query` and `params` as `client.fetch`, and switches from a
+   * GET to a POST for queries too large for the request URL in the same way,
+   * but queries the comments endpoint, which accepts none of the query options
+   * `client.fetch` does (`perspective`, `useCdn`, `filterResponse`,
+   * `resultSourceMap`, stega).
+   *
+   * Comment documents are of type `sanity.comment`, so queries typically filter
+   * on `_type == "sanity.comment"`.
```
pepahlavacek: mentioning this might be confusing - does it mean that it's filtered out for me already (no), or that i should filter it by the type == sanity.comment

src/collaboration/CollaborationCommentsClient.ts
```diff
@@ -0,0 +1,385 @@
...
+   *
+   * @param query - GROQ-filter to listen to changes for
+   * @param params - Optional query parameters
+   * @param options - Listener options, such as `events`, `includeResult` and `tag`
+   */
+  listen<Opts extends CollaborationCommentsListenOptions>(
+    query: string,
+    params: QueryParams | undefined,
+    options: Opts,
+  ): Observable<ListenEventFromOptions<CollaborationCommentDocument, Opts>>
```
pepahlavacek, line 202: do we actually have to type it twice? if we don't support most of the options we might as well strip them?

src/collaboration/types.ts
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * @alpha
+ */
+export interface CollaborationCommentUpdate {
+  /** Replaces the current message */
+  message?: CollaborationCommentMessage
+  /** Cascades to the comment's replies */
+  status?: CollaborationCommentStatus
+  /**
+   * Re-anchors the comment within the field it already targets.
```
pepahlavacek: within the field and sourceDocumentId it already targets

test/collaborationCommentsClient.test.ts
```diff
@@ -0,0 +1,1302 @@
...
```
pepahlavacek, line 434: i thinks i'm missing something - are we actually asserting anything here?

pepahlavacek (dismissed):
A few small comments I'll check on in the next round, but I think we're in a good shape

src/types.ts
hermanwikner: Already fixed in recent commit! 👍

pepahlavacek: Trimmed to a single line in 387a74a.

test/client/data.mutations.test.ts
pepahlavacek: Moot now: fb80c29 dropped the unrelated test changes from this PR entirely.

test/collaborationComments.test-d.ts
pepahlavacek: Agreed, removed. `resolves to the comment document` and `delete resolves to a mutation result` already assert both of these positively, so the `@ts-expect-error` pair only restated them.

src/collaboration/CollaborationCommentsClient.ts
pepahlavacek: Reworded so it reads as an instruction rather than a fact:

> The query runs against the organization store, which is not scoped to comments, so filter on `_type == "sanity.comment"`.

Same change in the README.

src/collaboration/CollaborationCommentsClient.ts
pepahlavacek: Kept both overloads. They are what makes `listen(query)` narrow to `MutationEvent<CollaborationCommentDocument>` rather than the full `ListenEvent` union, and it keeps the shape identical to how `client.listen` is typed in `src/data/listen.ts`.

On stripping the options: I went looking in the collaboration repo, and `/collaboration/comments/listen` is not implemented there at all. Kong strips `organizationId`, `resourceType` and `resourceId` and proxies everything else to the organization store listener, so `events`, `enableResume`, `includeResult`, `includeMutations`, `includePreviousRevision`, `visibility`, `effectFormat` and `tag` are all genuinely honored.

The one exception is `includeAllVersions`, which can never make a difference: comments are written directly as `sanity.comment` documents with no drafts or versions. That one is now omitted from `CollaborationCommentsListenOptions`.

Also replaced the `@param options` text, which listed three options as though they were the only ones.

src/collaboration/types.ts
pepahlavacek: Correct, and worth being explicit about: `updateComment.ts` reads both `target.path.field` and `target.sourceDocumentId` off the stored comment and resolves the new range against that document. Neither can be changed through the request. Updated the JSDoc and the README to name both.

test/collaborationCommentsClient.test.ts
pepahlavacek: It does assert, but only invisibly: `get-it/mock` throws on a request matching no registered handler, and `uninstallMock()` calls `assertAllConsumed()`, so a missing or wrong `transactionId` does fail the test. That is far too implicit to read, so there is now an explicit assertion on the recorded query params, matching the `uses resource query parameters` test above:

```ts
expect(
  getActiveMock()
    .getRequests()
    .map((request) => request.query.transactionId),
).toEqual(['txn-123', 'txn-123', 'txn-123', 'txn-123'])
```

`applies the request tag prefix on every write and on fetch`, right below, had the same problem and got the same treatment.

src/collaboration/CollaborationCommentsClient.ts
pepahlavacek: it's following the client.listen way, so keeping as is

src/types.ts
```diff
@@ -126,6 +126,10 @@ type ClientConfigResource =
 export interface ClientConfig {
   projectId?: string
   dataset?: string
+  /**
+   * Organization ID for organization-scoped APIs.
+   */
+  organizationId?: string
```
ryanbonial: Heads up. I think that this will [likely affect functions invocations](https://github.com/sanity-io/client/blob/edex-1050/src/functions/invoke.ts#L58), not sure if that is intentional. From what I am seeing setting an org ID to use comments will cause that client to always look at the org level when invoking functions which will likely be unexpected.

pepahlavacek: Great catch, thanks @ryanbonial!

Hey @ryanbethel, seeing you beat us to adding the orgId! I think that going forward there will be more features depending on it. We'll probably scope the collaboration specific one to `{ collaboration: { organizationId } }`, but wonder if we should either solve it by scoping it to every function

```
createClient({
  projectId: '<id>',
  dataset: '<dataset>',
  functions: {
    stackId: '<stack>',
    organizationId: '<org-for-functions>',
  },
  collaboration: {
    organizationId: '<org-for-comments>',
  },
})
```

or solve it with flags

```
createClient({
  projectId: '<id>',
  dataset: '<dataset>',
  organizationId: '<org-for-functions>',
  stackId: '<stack-id>',
  functions: {
    useOrgScope: true // or something prettier 
  },
})
```

..or if scoping functions to organizationId is the preferred way forward we might as well keep it as is :)

pepahlavacek (approved):

PR #1258, opened 2026-08-10
feat: RUN-1528 invoke function helper

ryanbethel:
## feat: `client.functions.invoke()`

Adds a client method for invoking a deployed Pubsub functions.

```ts
const client = createClient({
  projectId: 'abc123',
  apiVersion: '2025-02-19',
  token: '...',
  stackId: 'ST-1234567890',
})

const result = await client.functions.invoke<{words: number}>('summarize', {
  event: {data: {documentId: 'abc123'}},
})
```

## API
`client.functions.invoke<R>(name, request?): Promise<R | undefined>`
`client.observable.functions.invoke<R>(name, request?): Observable<R | undefined>`

### How name resolution works
The invoke route is keyed by function id, but callers know functions by the name declared in
the blueprint. Names are unique within a stack, so each call is two requests:

`GET /blueprints/stacks/:stackId` to find the resource whose type starts with `sanity.function.` and whose name matches, then read its `externalId`
`POST /functions/:functionId/invoke with {event: {data}}`
A per-call `stackId` overrides the config, so a single client can reach more than one stack.

---

> [!NOTE]
> **Medium Risk**
> New public API surface and authenticated remote execution paths, but behavior is isolated behind `client.functions` with validation before invoke and broad test coverage.
> 
> **Overview**
> Adds **`client.functions.invoke()`** (and **`client.observable.functions.invoke()`**) so callers can run deployed **`sanity.function.pubsub`** functions by blueprint **name**, with **`event.data`**, optional **`timeout`**, and **`AbortSignal`**.
> 
> Each call is two HTTP steps: **`GET /blueprints/stacks/:stackId`** to map the name to a function id (only pubsub types are allowed), then **`POST /functions/:id/invoke`**. New client config **`stackId`** and **`organizationId`** (per-call overrides supported) drive scope via **`X-Sanity-Scope-*`** headers; org scope can work without **`projectId`** when **`useProjectHostname: false`**.
> 
> README and a minor changeset document the API; vitest coverage exercises resolution, errors, scoping, and observable parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a2d3c94a459868cb230399c0116854b21b6270d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/functions/invoke.ts
+++ b/src/functions/invoke.ts
@@ -6,6 +6,7 @@ import type {HttpRequest, InitializedClientConfig} from '../types'
 
 /** Function resource types in a blueprint are namespaced under this prefix. */
 const FUNCTION_RESOURCE_PREFIX = 'sanity.function.'
+const INVOKABLE_FUNCTION_TYPE = 'sanity.function.pubsub'
 
 /** @public */
 export interface InvokeFunctionEvent {
@@ -129,6 +130,10 @@ function _resolveFunctionId(
         throw new Error(
           `Function "${functionName}" is declared in stack "${stackId}" but is not deployed`,
         )
+      }
+
+      if (match.type !== INVOKABLE_FUNCTION_TYPE) {
+        throw new Error(`Function invocation is not supported for ${match.type}`)
       }
 
       return match.externalId
--- a/src/functions/invoke.ts
+++ b/src/functions/invoke.ts
@@ -25,6 +25,10 @@ export interface InvokeFunctionRequest {
    */
   stackId?: string
   /**
+   * Organization owning the stack.
+   */
+  organizationId?: string
+  /**
    * Milliseconds to wait for the function to return.
    */
   timeout?: number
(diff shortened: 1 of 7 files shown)


src/functions/invoke.ts
```diff
@@ -0,0 +1,171 @@
...
+      if (!match) {
+        throw new Error(`Function "${functionName}" not found in stack "${stackId}"`)
+      }
+
+      if (!match.externalId) {
+        throw new Error(
+          `Function "${functionName}" is declared in stack "${stackId}" but is not deployed`,
+        )
+      }
+
```
macdonst, line 134: We need to add one more condition here so that if the function type is not `sanity.function.pubsub` then we need to throw an error along the lines of `Function invocation is not supported for ${match.type}`

This way we fail fast without having to make another call to the Functions service.

test/functions/invoke.test.ts
```diff
@@ -0,0 +1,316 @@
...
+
+const stackUri = (stackId: string) => `${BASE}/blueprints/stacks/${stackId}`
+const invokeUri = (functionId: string) => `${BASE}/functions/${functionId}/invoke`
+
+const STACK = {
+  id: STACK_ID,
+  name: 'my-stack',
+  resources: [
+    {name: 'my-func', type: 'sanity.function.document', externalId: 'fn-abc123'},
+    {name: 'other-func', type: 'sanity.function.queue', externalId: 'fn-def456'},
```
macdonst: Sorry, it looks like a lot of the tests need to get reworked because of the `sanity.function.pubsub` type. I'm not going to comment on every place below here that needs to be changed.

But we should add a test when the client tries to invoke a doc function and gets an error.

macdonst (changes requested):
Sorry to do this to you but the change to `pubsub` will necessitate some changes on this PR.

test/functions/invoke.test.ts
ryanbethel: Yeah, I cleaned those up. Thanks for catching it.

src/functions/invoke.ts
ryanbethel: I was relying on the API rejecting, but failing early is probably better.

src/functions/invoke.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (match.type !== INVOKABLE_FUNCTION_TYPE) {
```
macdonst, line 135: I think this is fine for now but eventually this will become an array with queues, durables.

macdonst (approved):
LGTM, I tested it in production.

PR #1260, opened 2026-08-11
build!: upgrade @sanity/pkg-utils to v12, swap eslint/prettier for oxlint/oxfmt

rexxars:
Upgrades `@sanity/pkg-utils` from v10.5 to v12.1, which forced two things along with it.

## Why the linter swap is in here

pkg-utils v12 requires TypeScript `6.x || 7.x` as a peer. There is no stable TS 6 (npm's `beta` tag is `6.0.0-beta`), so in practice it means TypeScript 7. typescript-eslint refuses TS 7 at require time:

```
Error: typescript-eslint does not support TS 7.0.
```

No typescript-eslint release supports it yet ([typescript-eslint#10940](https://github.com/typescript-eslint/typescript-eslint/issues/10940)), so there was no version combination that installed cleanly. Since we were planning to move to oxlint/oxfmt anyway, that swap happens here rather than pinning ourselves to `--legacy-peer-deps`. `npm install` now resolves with no flags, and dependency count drops from 764 to 604.

## Changes

- `@sanity/pkg-utils` 10.5 to 12.1, `typescript` 5.9.3 to 7.0.2
- `package.config.ts` migrated to the v12 tsdown build stack: `extract` becomes `tsdoc`, the `external` callback becomes `deps.alwaysBundle`, and `rollup.hashChunkFileNames` is dropped since chunks are content-hashed by default
- eslint and prettier plus 12 plugins replaced by oxlint and oxfmt. Import sorting moves from `eslint-plugin-simple-import-sort` to oxfmt's `sortImports`, and package.json sorting from `prettier-plugin-packagejson` to oxfmt's `sortPackageJson`. Disable directives are migrated to `oxlint-disable`, and three that no longer matched an enabled rule are removed
- `rimraf` and the `prebuild`/`clean` scripts replaced by `clean: ['dist', 'coverage']` in `package.config.ts`
- CI: `.github/eslint-compact.json` dropped in favour of oxlint's native `-f github` output, and `prettier.yml` renamed to `format.yml`

## One real regression this caught

v12 regenerates the `exports` map and emits only `import` under the `node` condition, dropping the catch-all:

```diff
 "node": {
   "source": "./src/index.node.ts",
   "import": "./dist/index.node.js",
-  "default": "./dist/index.node.js"
 },
-"import": "./dist/index.js",
 "default": "./dist/index.js"
```

`require(esm)` activates `require`, not `import`, so Node's resolution enters the `node` branch, finds nothing active, backtracks out, and falls through to the top-level `default`. CJS consumers would have silently got the platform-neutral fetch build, skipping the Node middleware: Readable body conversion, explicit proxy support, and the lineage/User-Agent headers.

`test/exports.test.ts` caught this, which is exactly what that guard exists for. Filed upstream as [sanity-io/pkg-utils#3280](https://github.com/sanity-io/pkg-utils/issues/3280), fixed by @stipsan in [pkg-utils#3282](https://github.com/sanity-io/pkg-utils/pull/3282) and released in 12.1.2, which this PR now depends on. The local `exports` resolver that worked around it has been removed.

Verified rather than assumed: resetting the `node` branch to the broken v12.1.0 shape and rebuilding regenerates `node.default` from upstream, and the guard test passes with no workaround in place.

## One deliberate concession

`ae-missing-release-tag` is downgraded from `error` to `warn`. The v12 dts pipeline synthesizes the `studioPath` namespace re-export in `csm` as `declare namespace studioPath_d_exports` and drops the `@alpha` tag. Tagging the import, the export, `export * as`, and the `tsgo` generator all produce the same result. Filed upstream as [sanity-io/pkg-utils#3281](https://github.com/sanity-io/pkg-utils/issues/3281).

Worth being precise about the blast radius, since it sounds worse than it is:

- The published types do not change. Building this same commit in a worktree with pkg-utils 10.9.2 also emits an untagged `declare namespace studioPath`. v12 emits the same untagged namespace under a synthesized name. The check got stricter about something that was already true
- The public API is intact. Verified by packing this branch and compiling a consumer against the tarball using `studioPath.Path` as a type and `studioPath.toString`/`fromString` as values
- What is actually lost is CI enforcement. Verified by adding an untagged export, which now warns and exits 0 where it previously failed the build. Every current export is tagged, so the cost today is zero and the exposure is future drift

If we want the gate back before upstream lands a fix, a CI step that fails on any `ae-missing-release-tag` occurrence other than `studioPath_d_exports` would restore it with a single named exception. Happy to add that here or as a follow-up.

## Deno

The v12 dts pipeline code-splits shared types into type-only chunks and references them from the emitted declarations with `.js` specifiers that have no JS counterpart. tsc substitutes the sibling `.d.ts`; Deno resolves the specifier literally and fails with `TS2307`, which broke `npm run test:deno`. Known upstream as [sanity-io/pkg-utils#3162](https://github.com/sanity-io/pkg-utils/issues/3162).

Deno only applies that substitution for packages resolved through `node_modules`. The test was reaching into `../../dist/index.js` by relative path, which no consumer does, so it now packs the build and installs it, resolving `@sanity/client` the way real consumers do. Type-checking stays on and the test covers the published artifact.

The stub-emitting workaround from #3162 was tried and reverted. It clears `TS2307`, but this package's public types use `Buffer` and `NodeJS.ReadableStream`, which then fail to resolve in the same context, and adding `/// <reference types="node" />` to the chunk made it worse by breaking Deno consumers of the published package outright. Deno consumers type-check cleanly with and without those stubs, so they only added files to `dist`. Findings are [posted upstream](https://github.com/sanity-io/pkg-utils/issues/3162#issuecomment-5257294561).

## Other notes

- Fixed an unsafe optional chain in `src/data/eventsource.ts` that the new linter flagged. `(event?.data as {message: string}).message` throws when `event` is null, so it now relies on the tuple's discriminated narrowing
- Fixed a pre-existing bug in the sibling branch of that same expression, caught by Bugbot on this PR. `MessageParseError` was built with `cause: event`, which is always `null` on the parse-failure path, discarding the actual error. It now passes `parseError`, matching what the `onMessage` path already does
- `CHANGELOG.md` is excluded from oxfmt, since its markdown output differs from prettier's and would fight release-please
- oxlint runs the `correctness` category only. `suspicious` was tried and produced a large volume of findings unrelated to this change, so it is left for a separate pass if wanted

## Verification

From a clean `npm ci`:

- `npm run build`: passes
- `npm run lint -- --report-unused-disable-directives -f github`: 0 warnings, 0 errors
- `npm run format:check`: clean
- `vitest run`: 39 files, 1092 passed, 2 skipped, no type errors

All CI checks pass, including every alternate-runtime suite (deno, bun, workerd, vercel-edge, browser, next, node-runtimes) and the ubuntu/macos/windows test matrix.

---

> [!NOTE]
> **High Risk**
> Major build/export pipeline and TypeScript upgrade that regenerates package exports and weakens a release-tag gate; incorrect exports nearly routed CJS consumers to the wrong Node build.
> 
> **Overview**
> **Breaking build-tooling upgrade** to `@sanity/pkg-utils` v12 and TypeScript 7. Replaces eslint/prettier (and ~12 plugins) with **oxlint/oxfmt** because typescript-eslint does not support TS 7 yet.
> 
> Migrates `package.config.ts` to the v12 API (`extract` → `tsdoc`, `deps.alwaysBundle`, built-in `clean`). CI drops the custom ESLint problem matcher and renames the Prettier workflow to Format.
> 
> Also fixes an unsafe optional chain and a wrong `cause` on `MessageParseError` in `eventsource.ts`, reworks the Deno test to install the packed package (avoiding a dts chunk-resolution issue), and downgrades `ae-missing-release-tag` to warn for a known v12 dts quirk around `studioPath`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 707c7f9a9476661a9af700b0acd35fafad06994a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rexxars:
@stipsan Would tsdown be a better choice than pkg-utils here? Thoughts? (given `get-it` and `@sanity/ui` etc went with tsdown directly)

src/data/eventsource.ts
```diff
@@ -152,7 +152,7 @@ function connectWithESInstance<EventTypeName extends string>(
         observer.error(
           parseError
             ? new MessageParseError('Unable to parse EventSource error message', {cause: event})
-            : new MessageError((event?.data as {message: string}).message, event),
+            : new MessageError((event.data as {message: string}).message, event),
```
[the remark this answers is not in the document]
rexxars: Good catch, confirmed and fixed in 97a354c.

`parseEvent` returns `[Error, null]` on failure, so `cause: event` was always `null` on that branch and the real parse failure was dropped. It now passes `parseError`, matching what the `onMessage` path at line 194 already did.

Worth noting for the record that this one is pre-existing on `main` rather than introduced here. This PR only removed the `?.` from the sibling branch of the same ternary, which is what put the expression in the diff.

package.json
```diff
@@ -36,58 +41,60 @@
...
-    "clean": "npx rimraf dist coverage",
     "coverage": "vitest run --coverage",
-    "lint": "eslint . --max-warnings 0",
+    "lint": "oxlint",
     "prepublishOnly": "npm run build",
     "test": "vitest",
     "test:browser": "vitest run --config ./vitest.browser.config.ts",
     "test:bun": "bun test runtimes/bun",
-    "test:deno": "deno test --allow-read --allow-net --allow-env --fail-fast --node-modules-dir=manual runtimes/deno",
+    "test:deno": "npm pack --pack-destination runtimes/deno && cd runtimes/deno && npm install --no-save --no-package-lock ./sanity-client-*.tgz @types/node && deno test --allow-read --allow-net --allow-env --fail-fast --node-modules-dir=manual .",
```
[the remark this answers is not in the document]
rexxars: Checked this against the actual run and it does not reproduce: the bare specifier resolves and the tests pass with type-checking on.

From the `deno-runtime` job log for 97a354c ([job 93887002448](https://github.com/sanity-io/client/actions/runs/31523726625/job/93887002448)):

```
> npm pack --pack-destination runtimes/deno && cd runtimes/deno && npm install --no-save --no-package-lock ./sanity-client-*.tgz @types/node && deno test --allow-read --allow-net --allow-env --fail-fast --node-modules-dir=manual .
sanity-client-7.26.2.tgz
added 12 packages, and audited 13 packages in 3s
Check client_fetch_test.mjs
running 2 tests from ./client_fetch_test.mjs
native fetch ... ok (336ms)
@sanity/client fetch ... ok (236ms)
ok | 2 passed | 0 failed (575ms)
```

The `Check client_fetch_test.mjs` line matters: type-checking ran, so this is not a green tick from `--no-check` or a skipped graph. Same result locally on Deno 2.7.12.

`--node-modules-dir=manual` tells Deno the directory is externally managed and to read it as-is, so bare-specifier resolution walks `node_modules` directly. It is `auto` mode that reconciles against declared dependencies and installs what is missing. `npm install --no-save` writes a complete package layout into `node_modules`; it only skips recording the entry in `package.json`.

There is a real constraint nearby, and it is why the install line looks like this. Deno *does* require declared packages for **type reference directives**, which is a separate resolution path. That failure looks like:

```
Could not find a matching package for 'npm:@types/node' in the node_modules directory.
Ensure you have all your JSR and npm dependencies listed in your deno.json or package.json
```

I hit exactly that while developing this, which is why `@types/node` is installed explicitly alongside the tarball rather than relied on from the root workspace.

Conceding one point: depending on an unlisted package does lean on manual-mode behaviour that Deno could tighten later. The alternatives are worse. Dropping `--no-save` leaves the tracked `runtimes/deno/package.json` dirty after every run, and a hardcoded `file:` dependency pins the version string and breaks on every release. If Deno ever changes this, CI fails loudly rather than silently degrading.

.oxlintrc.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$schema": "./node_modules/oxlint/configuration_schema.json",
+  "plugins": ["typescript", "unicorn", "oxc", "import"],
+  "categories": {
+    "correctness": "error"
+  },
+  "ignorePatterns": ["dist", "coverage", "runtimes/deno", "runtimes/node"],
```
stipsan, line 7: I would also set these options in a follow-up:
```suggestion
  "plugins": ["typescript", "unicorn", "oxc", "import"],
  "options": {
     "typeAware": true,
     "reportUnusedDisableDirectives": "error"
  },
  "categories": {
    "correctness": "error"
  },
```
There are many incredibly useful `--type-aware` rules, and it's also nice to ban disable directives that no longer have an effect

package.config.ts
```diff
@@ -3,15 +3,33 @@ import {defineConfig} from '@sanity/pkg-utils'
...
+  // Guarded by test/exports.test.ts.
+  exports: (prev) =>
+    Object.fromEntries(
+      Object.entries(prev).map(([path, entry]) => [
+        path,
+        entry.node?.import && !entry.node.require
+          ? {...entry, node: {...entry.node, require: entry.node.import}}
+          : entry,
+      ]),
+    ),
```
stipsan: Fixed in https://github.com/sanity-io/pkg-utils/releases/tag/%40sanity%2Fpkg-utils%4012.1.2

stipsan (dismissed):

package.config.ts
rexxars: Upgraded to 12.1.2 and removed the workaround in 3197a0f. Thanks for turning that around so fast.

Verified rather than assumed: I reset the `node` branch back to the broken v12.1.0 shape (`{source, import}`), rebuilt, and pkg-utils regenerated it as

```json
"node": {
  "source": "./src/index.node.ts",
  "import": "./dist/index.node.js",
  "default": "./dist/index.node.js"
}
```

`test/exports.test.ts` passes with the resolver gone, so the guard is now asserting against upstream behaviour rather than our patch.

.oxlintrc.json
rexxars: Agreed on both, and agreed they belong in a follow-up rather than here. Will do another cleanup pass for this later.

stipsan: I agree `no-base-to-string` and `restrict-template-expressions` aren't very useful. In other codebases `no-redundant-type-constituents` (and `no-duplicate-type-constituents`, `no-unnecessary-type-parameters` and other `suspicious` category rules) have lead to a nice cleanup in the code. `no-floating-promises` and `await-thenable` often helped reduce flake in test suites, and in some cases caught real bugs.<br>The `no-deprecated` rule has also been helpful, though I expect it to be annoying in this codebase since we use it to mark APIs as "internal"

.github/workflows/ci.yml
```diff
@@ -36,9 +36,7 @@ jobs:
       # engines conflicts from dev-only deps (e.g. vitest).
       - run: npx ls-engines --mode virtual --no-dev
       - run: npm run prepublishOnly
-      - name: Register Problem Matcher for ESLint that handles --report-unused-disable-directives
-        run: echo "::add-matcher::.github/eslint-compact.json"
-      - run: npm run lint -- --report-unused-disable-directives -f compact
+      - run: npm run lint -- --report-unused-disable-directives -f github
```
stipsan: `-f github` is detected automatically

package.json
```diff
@@ -3,27 +3,32 @@
...
   "repository": {
     "type": "git",
     "url": "git+https://github.com/sanity-io/client.git"
   },
-  "license": "MIT",
-  "author": "Sanity.io <hello@sanity.io>",
-  "sideEffects": false,
+  "files": [
+    "dist",
+    "src"
```
stipsan, line 27: Note for follow-up, we can shrink the npm tarball size quite a bit if we stop publishing the `src` folder

.github/workflows/ci.yml
rexxars: Confirmed and dropped in 707c7f9.

Amusing detour worth recording: my first attempt to verify this said you were wrong. `GITHUB_ACTIONS=true npx oxlint …` kept printing the default format, through several env combinations. It only reproduced in a clean `env -i`:

```
$ env -i PATH=$PATH HOME=$HOME GITHUB_ACTIONS=true ./node_modules/.bin/oxlint -D no-console /tmp/probe.ts
::error file=/tmp/probe.ts,line=2,endLine=2,col=1,endColumn=12,title=eslint(no-console)::…
```

Bisecting the difference, the variable suppressing it is `CLAUDECODE=1`. oxlint ships an `agent` formatter and prefers it when it detects an agent environment, which outranks the GitHub detection. So the flag is redundant in CI exactly as you said, and my local disagreement was an artifact of running inside Claude Code.

package.json
rexxars: Measured it so the follow-up has numbers: **484 KB → 388 KB**, so about 96 KB / 20% off the tarball. Unpacked it is 620 KB of `src` against 1.9 MB of `dist`.

- `publishConfig.exports` contains no `/src/` references, so the `source` condition is already stripped from the published map and nothing in it would dangle.
- The sourcemaps embed `sourcesContent` in full, so debugging does not depend on `src` shipping alongside. The published source is currently duplicated: once as `src/*.ts`, once inside `dist/*.map`.

Which suggests the bigger win might be the maps rather than `src`. Worth deciding in the follow-up whether to keep shipping both.

.oxlintrc.json
rexxars: That matches what I saw. Revised follow-up scope based on your read:

- **In**: `no-redundant-type-constituents`, `no-duplicate-type-constituents`, `no-unnecessary-type-parameters`, `no-floating-promises`, `await-thenable`, plus `reportUnusedDisableDirectives`
- **Out**: `no-base-to-string`, `restrict-template-expressions`. Nearly all 12 hits were `received ${value}` in validation error messages, so the payoff is `[object Object]` becoming a slightly better string in a message the user only sees when they already passed the wrong type

Agreed on `no-deprecated` being awkward here. `@deprecated` is load-bearing for the internal-API convention in this codebase, `ReleaseId` and `ReplaceDraftAction` being the obvious examples, so it would fire on intentional usage. Probably not worth it unless the internal markers move to a different tag.

Worth flagging on the `suspicious` category: I tried enabling it wholesale early on and backed it out. The volume was dominated by `no-underscore-dangle` firing on `_id`, `_key`, `_type` and friends, which is just the document shape rather than anything to clean up. It would need that rule disabled to be usable, after which the type-constituent rules you mention are the interesting part.

package.json
stipsan: We should keep shipping maps, they are useful for debugging. But `src` have very little value beyond creating bloat

stipsan (approved):

PR #1261, opened 2026-08-11
chore: switch releases to changesets, migrate to pnpm

rexxars:
Replaces release-please with changesets, and migrates the repo from npm to pnpm so it matches the rest of our tooling.

It also sets up the `v7` maintenance line alongside `main`, so we can keep shipping 7.x bugfixes after v8 goes out.

## Keeping the two release lines apart

Only the current major may take the `latest` dist-tag on npm and the "Latest" badge on GitHub releases. Otherwise `npm install @sanity/client` resolves to an old major every time a 7.x patch ships. `release.yml` derives both from the branch it runs on:

- `main` publishes `latest` and creates its GitHub release with `--latest=true`.
- `v7` publishes `latest-v7` and creates its GitHub release with `--latest=false`.

Two things worth calling out:

- The dist-tag is `latest-v7`, not `v7`. npm rejects a dist-tag that parses as a valid semver range, and `v7`, `7`, and `7.x` all parse as `>=7.0.0 <8.0.0`.
- `changesets/action` calls `repos.createRelease` without `make_latest`, so GitHub marks whatever was released most recently as "Latest". A 7.x patch would steal the badge from v8. Demoting it afterwards does not promote the previous release back, so it has to be right at creation time. The workflow therefore runs with `createGithubReleases: false` and creates the release itself. That also stops the action from pushing the git tag, so the workflow pushes it first, and `.github/scripts/release-notes.mjs` extracts the changelog section the action would have used.

## What contributors need to do

Nothing new. `generate-changeset.yml` writes the changeset from the PR title, the same as in get-it, so a conventional-commit title is still the only requirement. Manual changesets via `pnpm changeset` are respected and never overwritten.

## v8

The breaking change in #1217 merged before changesets existed, so nothing would have triggered a major bump. `.changeset/pr-1217.md` records it as `major`, carrying the `BREAKING CHANGE:` notes from that commit. Verified that `changeset version` on `main` produces 8.0.0.

`CHANGELOG.md` loses its multi-line preamble. Changesets inserts new entries after the first line, so with the old release-please header new releases landed above the title.

## pnpm notes

- `pnpm run x -- --flag` forwards a literal `--` to the script, which breaks downstream arg parsing. The CI invocations are written without it.
- `ls-engines --mode virtual` cannot work under pnpm: it reads lockfiles through arborist, which only understands `package-lock.json`. CI uses `--mode actual --no-dev`, which still prunes dev deps.
- The `prod-deps` job is gone. A pnpm `node_modules` is a symlink farm into the store and does not survive being cached across jobs, so the deno, bun, and node-runtime jobs each run `pnpm install --prod` directly.
- `pnpm/action-setup` is on v6 everywhere. Publishing uses OIDC trusted publishing, and `changeset publish` reaches npm via `pnpm publish` spawning `npm publish`. The OIDC token only survives that chain on pnpm >= 11.0.9 with action-setup >= v6.0.6 (pnpm/pnpm#11566). On v4 the first publish fails with a masked 404. The shared org workflow was on v4 too; that is fixed in sanity-io/.github#29.

## Verified locally

Install, build, lint, prettier, and vitest (1092 passed), plus the node-runtimes, bun, deno, vercel-edge, workerd, and next suites. `changeset version` was run end to end and reverted. The browser suite was not run locally, since it needs a Playwright browser download.

## Follow-up

Once this lands, cherry-pick it onto `v7` and change only `baseBranch` to `v7` in `.changeset/config.json`. The workflows are branch-aware so both branches can hold identical files.

Filed upstream as sanity-io/.github#38: the shared `changesets.yml` has no dist-tag or non-latest-release support, so it cannot publish a maintenance branch. The `action-setup@v4` half of this has since been fixed in sanity-io/.github#29.

---

> [!NOTE]
> **High Risk**
> Rewrites the npm publish and GitHub release pipeline (OIDC, dist-tags, dual release lines), so mistakes can mis-tag releases or publish the wrong major as `latest`. Also changes the package manager across all CI.
> 
> **Overview**
> **Replaces release-please with changesets** and migrates the repo from npm to pnpm. Also wires up a `v7` maintenance line alongside `main`.
> 
> **Release pipeline:** Removes `release-please.yml` and adds `release.yml` using `changesets/action`. `main` publishes under the `latest` npm tag with GitHub "Latest"; `v7` publishes as `latest-v7` without that badge. GitHub releases are created manually (with `release-notes.mjs`) so maintenance patches cannot steal the Latest badge.
> 
> **Auto-changesets:** `generate-changeset.yml` writes `.changeset/pr-<n>.md` from conventional-commit PR titles. Manual changesets win and are never overwritten. Seeds a major bump via `.changeset/pr-1217.md` so the next version on `main` is 8.0.0.
> 
> **pnpm migration:** Switches all workflows to `pnpm/action-setup@v6` and `pnpm install`. Drops the cached `prod-deps` job (pnpm's store layout does not cache across jobs). Adjusts Node matrix installs so the library floor (22.12) is still tested even though pnpm 11 needs 22.13+.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61e69feaa23426c0cd2b834fbe8abea420a48bee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CHANGELOG.md
```diff
@@ -1,10 +1,5 @@
-<!-- markdownlint-disable --><!-- textlint-disable -->
-
 # 📓 Changelog
```
stipsan: Changesets don't recognize this heading, it uses the package name, or just `# Changelog` (without the `📓 ` emoji), I recommend using the package name

```suggestion
# @sanity/client
```

stipsan (dismissed):
Changesets! Nice!!
Also nice to see `pnpm` so that the `publishConfig.exports` pattern can be used 😌

stipsan (dismissed):

rexxars:
Had to rebase due to some conflicts, sorry for the reboop request.

stipsan (approved):

PR #1276, opened 2026-08-13
feat: add client.context with knowledge bases, builds, triage, telemetry, and GROQ reads

RostiMelk:
Adds a `client.context` namespace for Sanity Context: knowledge bases, builds, ingest, curation, conversation telemetry, and GROQ reads over every Context document, in one place.

The shape follows the Context API's tiering (documents are queried, the control plane is called): every read is GROQ, either free-form or through a canned helper, and every verb is a typed REST method mirroring the published 30-op contract.

```ts
const client = createClient({..., context: {organizationId, resource: {type: 'knowledge-base', id}}})

// Reads: free-form GROQ over any Context document type
await client.context.fetch(groq, params)
client.context.listen(groq, params)

// ...or canned helpers (documented queries, identity + pagination params only)
await client.context.entries.get({path: 'pricing/plans'})
await client.context.issues.list({status: 'open'})

// The automation lifecycle
await client.context.imports.create({type: 'text', title: 'Refund policy', content: md})
const {jobId} = await client.context.build()
await client.context.jobs.get({jobId})

// Robot triage
await client.context.issues.resolve({issueId, resolution: 'accept_new'})

// Conversation telemetry
await client.context.conversations.save(threadId, {messages, metadata})
```

Surface: KB CRUD, build/cancelBuild/refresh + jobs polling, imports (create/list/get/download/delete, file uploads via signed URL), sources (list/get/content/delete), issue transitions (apply/resolve/dismiss/reopen), instruction writes, entries.rebuild, conversations save/classify, plus the eight GROQ read helpers typed from the spec's published doc components (EntryDoc, IssueDoc, InstructionDoc, McpDoc, ConversationDoc). Everything exports under the single `Context` type namespace.

The vendored spec is the Context API's published contract (sanity-io/atlas#385 + #387); reads deliberately have no REST twin, so the helper doc comments show the exact GROQ they run.

Closes SAGE-817, part of SAGE-985

---

> [!NOTE]
> **Medium Risk**
> Large new beta surface with org-scoped data access, async jobs, and direct signed-URL uploads; mostly additive but mistakes in triage/build calls could affect knowledge-base content.
> 
> **Overview**
> Adds a **beta** `client.context` API on both promise and observable clients, wired like other resource-scoped namespaces (e.g. media libraries).
> 
> **Configuration:** `resource: { type: 'knowledge-base', id }` scopes KB operations; `context.organizationId` scopes org-level GROQ and conversation telemetry. Asset uploads explicitly reject knowledge-base resources and point callers to `context.imports`.
> 
> **Reads:** `context.fetch` / `context.listen` hit the org Context document store (GET/POST query + EventSource listen). Canned helpers (`entries`, `issues`, `instructions`, `mcpEndpoints`, `conversations.get`) drain paginated GROQ internally.
> 
> **Writes (REST):** Knowledge base collection CRUD; per-KB build/cancel/refresh; imports (text/crawl/dataset/file with staged signed-URL upload); jobs; issue triage; instructions; entries rebuild; sources; conversation save/classify.
> 
> Types ship as `Context` from OpenAPI-derived `types.gen`; `_observe` is exported for observable wrappers. Minor version bump in changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93dfef0b0ebbe004b0aac09c877595491dcb63e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/context/ContextClient.ts
+++ b/src/context/ContextClient.ts
@@ -32,7 +32,6 @@ import type {
   CreateImportParams,
   CreateInstructionParams,
   CreateKnowledgeBaseParams,
-  DeleteSourceResponse,
   DismissIssueResponse,
   EditCrawlOptionsParams,
   EditInstructionParams,
@@ -422,11 +421,12 @@ export class KnowledgeBaseHandle {
         }),
         ...options,
       }),
-    delete: (params: {sourceId: string}, options?: RequestOptions) =>
-      this.#request<DeleteSourceResponse>(`/sources/${params.sourceId}`, {
+    delete: async (params: {sourceId: string}, options?: RequestOptions) => {
+      await this.#request<void>(`/sources/${params.sourceId}`, {
         method: 'DELETE',
         ...options,
-      }),
+      })
+    },
   }
 
   /** The audit feed: who did what, when. */
--- a/src/context/ContextClient.ts
+++ b/src/context/ContextClient.ts
@@ -49,7 +49,30 @@ import type {CreateFileImportParams, RenderFormat, RequestOptions} from './types
 
 type ListOptions = RequestOptions & {cursor?: string; limit?: number}
 
+type Client = SanityClient | ObservableSanityClient
+
 const COLLECTION_URL = '/context/knowledge-bases'
+
+/**
+ * The knowledge base every scoped method operates on comes from the client's
+ * `resource` configuration, matching how media libraries and canvases are
+ * addressed. Resolved per call so `withConfig` clones behave.
+ */
+function _resolveKnowledgeBaseId(client: Client): string {
+  const resource = client.config().resource
+
+  if (resource?.type !== 'knowledge-base') {
+    throw new Error(
+      '`resource` of type `knowledge-base` must be configured to use knowledge-base methods',
+    )
+  }
+
+  return resource.id
+}
+
+function _knowledgeBaseUrl(knowledgeBaseId: string, suffix = ''): string {
+  return `${COLLECTION_URL}/${encodeURIComponent(knowledgeBaseId)}${suffix}`
+}
 
 /** Serialize defined values into query params, dropping the undefined ones. */
 function _query(entries: Record<string, string | number | undefined>): Record<string, string> {
@@ -68,27 +91,43 @@ function _fetchBody(file: CreateFileImportParams['file']) {
 }
 
 /**
- * Everything scoped to one knowledge base, addressed by its id alone.
- * Construct via {@link ContextClient.knowledgeBase}.
+ * `client.context` — knowledge bases and everything scoped to them.
+ *
+ * Collection-level management (create, list, get, edit, delete) addresses
+ * knowledge bases per call, like `client.projects`. Everything scoped to one
+ * knowledge base (imports, builds, issues, entries, ...) operates on the
+ * client's configured `resource`, like media libraries:
+ *
+ * @example Full lifecycle
+ * ```ts
+ * const created = await client.context.knowledgeBases.create({
+ *   organizationId: 'org123',
+ *   title: 'Support docs',
+ *   description: 'Product docs and troubleshooting guides',
+ * })
+ *
+ * const kb = createClient({
+ *   apiVersion: '2026-08-25',
+ *   token,
+ *   resource: {type: 'knowledge-base', id: created.publicId},
+ * })
+ *
+ * await kb.context.imports.create({type: 'text', title: 'Refund policy', content: refundMd})
+ * const {jobId} = await kb.context.build()
+ * ```
  *
  * @beta
  */
-export class KnowledgeBaseHandle {
+export class ContextClient {
   #client: SanityClient
   #httpRequest: HttpRequest
-  #knowledgeBaseId: string
 
-  constructor(client: SanityClient, httpRequest: HttpRequest, knowledgeBaseId: string) {
+  constructor(client: SanityClient, httpRequest: HttpRequest) {
     this.#client = client
     this.#httpRequest = httpRequest
-    this.#knowledgeBaseId = knowledgeBaseId
   }
 
-  /** The knowledge base id this handle was constructed with. */
-  get knowledgeBaseId(): string {
-    return this.#knowledgeBaseId
-  }
-
+  /** Request against the configured knowledge base. */
   #request<R>(
     suffix: string,
     reqOptions: {
@@ -98,12 +137,12 @@ export class KnowledgeBaseHandle {
     } & RequestOptions = {},
   ): Promise<R> {
     return _request<R>(this.#client, this.#httpRequest, {
-      url: `${COLLECTION_URL}/${this.#knowledgeBaseId}${suffix}`,
+      url: _knowledgeBaseUrl(_resolveKnowledgeBaseId(this.#client), suffix),
       ...reqOptions,
     })
   }
 
-  /** Shared shape of every paginated list endpoint under the handle. */
+  /** Shared shape of every paginated list endpoint scoped to the knowledge base. */
   #list<R>(
     suffix: string,
     params?: ListOptions,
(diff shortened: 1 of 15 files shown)


src/context/ContextClient.ts
```diff
@@ -0,0 +1,576 @@
...
+  /** The built outline: every entry, ordered, with stats. */
+  outline(
+    params?: {format?: 'json' | 'markdown'},
+    options?: ContextRequestOptions,
+  ): Promise<ContextOutline> {
+    return this.#request('/outline', {
+      query: params?.format ? {format: params.format} : {},
+      ...options,
+    })
+  }
```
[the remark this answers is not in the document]
RostiMelk: Fixed in 307f49a. outline and entries.get now type by format: json resolves to the object shape, markdown and plain resolve to a string. Also picked up the plain format both endpoints support.

src/context/ContextClient.ts
```diff
@@ -0,0 +1,576 @@
...
+    ).then(
+      (kb) => ({organizationId: kb.organizationId, slug: kb.slug}),
+      (err) => {
+        // A failed resolution must not poison the handle for retries.
+        this.#address = undefined
+        throw err
+      },
+    )
+    return this.#address
+  }
```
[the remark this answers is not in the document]
RostiMelk: Fixed in 307f49a. The shared resolution no longer takes a caller's signal, so an abort can't reject concurrent callers awaiting the cached promise. The aborted caller still rejects: its own follow-up request starts with the aborted signal. Regression test added.

src/context/ContextClient.ts
```diff
@@ -0,0 +1,576 @@
...
+    const putResponse = await fetch(staged.uploadUrl, {
+      method: 'PUT',
+      // Buffer is typed as Uint8Array<ArrayBufferLike>, which BodyInit
+      // rejects even though fetch accepts it at runtime.
+      body: params.file as BodyInit,
+      ...(params.contentType && {
+        headers: {'content-type': params.contentType},
+      }),
+      signal: options?.signal,
+    })
```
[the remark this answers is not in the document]
RostiMelk: Fixed in 307f49a. The PUT now goes through config.resolveFetch with the configured proxy, falling back to global fetch. Tests inject the fetch through resolveFetch, same as the transport tests.

src/context/types.ts
rexxars, line 1: We're exposing a huge amount of types here, and blindly exporting all of them at the top level - I'm really not loving the idea of rapidly expanding our types footprint so significantly, especially when a lot of these are not properly namespaced - `ApplyIssuesParams` does not make it obvious that they're about Context/knowledge bases, as an example.

Can we rethink their names and export them under a common root, eg:

```ts
// contextTypes.ts
export type CrawlOptions = {…}
export type Changes = {…}

// types.ts
export * as Context from './context/contextTypes'
```

Then from an outside consumer, you'd refer to them as:

```ts
import {createClient, Context} from '@sanity/client'

const changes = Context.Changes = {…}
```

rexxars (changes requested):

src/context/types.ts
RostiMelk: Agreed, done in 959a3c3. Everything now exports under a single Context namespace exactly as you sketched: import {Context} from '@sanity/client', then Context.KnowledgeBase, Context.Changes, Context.ApplyIssuesParams. Dropped the Context prefix from the individual names since the namespace carries it. The package surface gains one top-level export instead of fifty.

src/context/ContextClient.ts
```diff
@@ -0,0 +1,608 @@
...
+          endLine: params.endLine,
+        }),
+        ...options,
+      }),
+    delete: async (params: {sourceId: string}, options?: ContextRequestOptions) => {
+      await this.#request<void>(`/sources/${params.sourceId}`, {
+        method: 'DELETE',
+        ...options,
+      })
+    },
```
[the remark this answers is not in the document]
rexxars: Do you want to address this, @RostiMelk ? Seems like an easy win for completeness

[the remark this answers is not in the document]
RostiMelk: Yep! I'm currently in the process of moving knowledge bases into their own resource, away from project datasets. Will do a pass asap

src/context/ContextClient.ts
```diff
@@ -0,0 +1,580 @@
...
+   */
+  async contentClient(options?: RequestOptions): Promise<SanityClient> {
+    const kb = await this.get(options)
+    if (!kb.sanityProjectId || !kb.sanityDatasetId) {
+      throw new Error(`Knowledge base "${this.#knowledgeBaseId}" has no dataset binding`)
+    }
+    return this.#client.withConfig({
+      projectId: kb.sanityProjectId,
+      dataset: kb.sanityDatasetId,
+      useCdn: false,
```
rexxars: Edge-case, but in case the original client is configured with a `resource` - lets clear it explicitly.

```suggestion
      useCdn: false,
      resource: undefined,
      '~experimental_resource': undefined,
```

src/context/ContextClient.ts
```diff
@@ -0,0 +1,580 @@
...
+        filename: params.filename,
+        ...(params.contentType && {contentType: params.contentType}),
+      },
+      ...options,
+    })
+    // The signed URL PUT goes straight to storage, outside the API pipeline:
+    // no auth header, and the body is raw bytes rather than JSON. Still uses
+    // the client's fetch resolution so proxy config applies in Node.
+    const config = this.#client.config()
+    const doFetch = config.resolveFetch?.(config.proxy) ?? (fetch as FetchFunction)
```
rexxars: I don't know that this needs a cast (contribution guidelines say no casts, need to update the lint rules to match)


```suggestion
    const doFetch: FetchFunction = config.resolveFetch?.(config.proxy) ?? globalThis.fetch
```

src/context/ContextClient.ts
```diff
@@ -0,0 +1,580 @@
...
+    const putResponse = await doFetch(staged.uploadUrl, {
+      method: 'PUT',
+      // Buffer's ArrayBufferLike generic misses the body type's ArrayBuffer
+      // bound, even though fetch accepts it at runtime.
+      body: params.file as FetchBody,
```
rexxars: We can handle this more gracefully without casting:

```suggestion
      body: _fetchBody(params.file),
```

…and add this:

```ts
function _fetchBody(file: CreateFileImportParams['file']) {
  if (!ArrayBuffer.isView(file)) return file
  return file.buffer instanceof ArrayBuffer
    ? new Uint8Array(file.buffer, file.byteOffset, file.byteLength)
    : Uint8Array.from(file)
}
```

rexxars (changes requested):

RostiMelk:
Dropped contentClient() from this PR before it ships as public API: knowledge base content is moving out of customer datasets into the per-org store (SAGE-889, merged), so a client bound to the KB's dataset would return a client pointed at nothing. Content reads go through the kb.entries / kb.issues surface this PR already provides; the dataset binding fields themselves leave the wire contract next (SAGE-892).

rexxars:
Addressed my two objections/suggestions, let me know if there was a reason for not implementing.

rexxars (dismissed):

RostiMelk:
Rebased onto the refreshed spec (see #1274 for the full API delta). Impact on this namespace:

- One real break, fixed: `sources.delete` now resolves to `void` — the API returns 204 No Content instead of 202 + `{jobId}`, so the review-round change that surfaced the job is reverted (the job no longer exists) and `DeleteSourceResponse` is removed
- KB addressing unchanged: the by-id resolve + org-scoped slug routes this PR is built on still exist with the same fields
- The new MCP/conversations/insights surface is present in the generated types but deliberately not exposed here — follow-up namespaces, keeping this PR its approved scope

All gates green on the rebased branch.

src/context/ContextClient.ts
RostiMelk: Applied in bc49147 (survives the rebase).

src/context/ContextClient.ts
RostiMelk: Applied as suggested in bc49147.

src/context/ContextClient.ts
```diff
@@ -0,0 +1,566 @@
...
+      method: 'POST',
+      body: {},
+      ...options,
+    })
+  }
+
+  /** Fetch the knowledge base. */
+  get(options?: RequestOptions): Promise<KnowledgeBase> {
+    return _getKnowledgeBaseById(this.#client, this.#httpRequest, this.#knowledgeBaseId, options)
+  }
```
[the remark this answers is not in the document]
RostiMelk: Fixed in c9d1688: get() now seeds the shared address cache, with a test covering the follow-up scoped call.

src/context/ContextClient.ts
```diff
@@ -0,0 +1,566 @@
...
+
+  /** Cancel the running build, if any. */
+  cancelBuild(options?: RequestOptions): Promise<{cancelled: boolean}> {
+    return this.#request('/build/cancel', {method: 'POST', ...options})
+  }
+
+  /** Run an incremental refresh: re-check sources and apply what changed. */
+  refresh(options?: RequestOptions): Promise<{jobId: string | null; started: boolean}> {
+    return this.#request('/refresh', {method: 'POST', ...options})
+  }
```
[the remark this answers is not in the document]
RostiMelk: Fixed in c9d1688: jobId is now string, matching the spec (the API always returns the in-flight job id when started is false; verified against the backend).

rexxars (dismissed):

RostiMelk:
Quick update @rexxars:

The Context API  has changed shape: reading documents (entries, issues, instructions, endpoints, conversations) now happens with GROQ queries instead of REST endpoints. This PR now follows that

So, removed: all the read methods whose endpoints no longer exist (entries, issues, instructions, outline, activity, revisions, crawl preview).

Added: `context.fetch` and `context.listen` for free-form GROQ, plus eight helpers for the common reads (`entries.list`, `issues.get`, and friends) so you don't have to hand-write a query for the basics. Each helper documents the exact query it runs.

Earlier feedback is in: no casts, one `Context` export, fetch through `resolveFetch`, observable parity 👍

src/context/ContextClient.ts
```diff
@@ -0,0 +1,991 @@
...
+  RebuildEntryResponse,
+  ReopenIssueResponse,
+  ResolveIssueParams,
+  ResolveIssueResponse,
+  SourceContentResponse,
+  SourceDetail,
+  SourcesResponse,
+  StagedUpload,
+} from './types'
+import type {
```
rexxars, line 52: Minor, but couldn't these import statements be merged?

rexxars (approved):
Cursor's feedback seems interesting to address in a follow-up PR if valid, but doesn't feel like it should block

PR #1282, opened 2026-08-15
refactor: clean up test suite, fix discovered issues

rexxars:
### Description

> [!CAUTION]
> > **_But Espen,_**
> 
> you say, starting a sentence with a conjunction,
> > **_have you lost your mind? 132 changed files, 14,507 additions and 11,082 deletions???
> > I am **_NOT_** reviewing that!_**

_Fair_ criticism! Let me defend myself!

The current test suite is a bit of a monstrosity. `client.test.ts` is almost 8000 lines long - a single file! And that's just one part of the test suite! You want more reasons? I've got em:

- The suite itself was cruft: one sprawling test file, environment guards scattered inline, no written conventions.
- We weren't actually testing in browsers. `test:browser` ran happy-dom; the job named `browser-runtime` started zero browsers. Now it runs real chromium, firefox and webkit.
- happy-dom stays, but as its own suite. People might use it to test their apps that rely on the client, so knowing where it breaks is still helpful.
- We weren't testing "edge" (vercel edge workers) or workerd (cloudflare workers) either. Half the edge suite was skipped by guards left over from a mocking library we stopped using years ago.
- Bun and Deno ran a two-line smoke file - they didn't run the _tests_. Every environment now runs the whole suite unless a filename opts out.
- Once the environments ran for real, they found bugs we had already shipped. Those are fixed here. Yay!
- Some code paths were untestable with mocks at all, so they were never tested. They are now.
- Mocks written from the implementation confirmed the _implementation,_ bugs included. A mock cannot falsify the thing it was copied from.
- So: a small, real-API smoke suite runs on every PR: one test per feature, across five runtimes.
- Everything else stays "offline", through one transport seam (get-it) with no _module_ mocking - we now use `get-it/mock` throughout, except for a few minor places (XHR browser upload path).
- Several of our checks were not checking anything, including a coverage flag that silently did nothing.
- We've now split test files by concern, with "what runs where" visible in the filename and the rules in CONTRIBUTING for reviewers to enforce.
- Costs: yeah, a bit more CI time, PRs now depend on the real API, and fork PRs get no integration coverage by default - _but_ - I feel way more confident about making changes.

**Aside from the test suite being a ton better (IMHO), these fixes also landed (as a result of our findings):**

- The live content API CORS probe now goes through the configured transport instead of `fetch` directly, ensuring it adheres to any proxies/middleware etc.
- `client.delete(id)` and `client.mutate(mutations)` are now typed as resolving to a mutation result rather than a document, so reading `._id` off the result is caught by `tsc` instead of silently being `undefined` at runtime.
- Server-sent events now carry their event IDs on Cloudflare Workers, so `listen()` and `live.events()` can resume from where they left off instead of reconnecting with an empty `Last-Event-ID`.
- `client.assets.upload()` against a Media Library now resolves to the uploaded asset instead of `undefined`, since that endpoint responds with `{asset}` where Content Lake responds with `{document}`.
- The `createVersion()` warning no longer suggests `baseId` when you're creating a document that doesn't exist yet, where `baseId` isn't a valid option in the first place.
- Documented that `listen()` and `live.events()` need a Cloudflare Workers `compatibility_date` of `2024-11-11` or later, since earlier dates reject the `cache` field that server-sent events set.

Sorry about not splitting these out into separate PRs, but I'm exhausted 🙃

### What to review

Honestly, don't expect anyone to review this super closely. Have a look at the diffs in `src` - that's 364 lines of changes across 13 files, so should be doable. Those are bugfixes found by the new test suite! Epic, eh? I think so.

Changes in `test` are.. huge. But; most of it is _relocated_ rather than new - split into focused files, cleaned up a bit. Don't expect you to read through these, but maybe have a glance at the patterns? The cool thing is you should be able to run the test suite(s) and tell from the coverage report that we're still doing a good job ;)

### Testing

Let them run! Both in CI and locally.

---

> [!NOTE]
> **Medium Risk**
> Several user-visible client fixes (SSE IDs, Media Library uploads, CORS probe transport) plus stricter TypeScript on `delete`/`mutate` defaults; the bulk is test/CI but integration tests now depend on secrets and real API on main-repo PRs.
> 
> **Overview**
> **Testing and CI** are reworked so the full Vitest suite runs across Node (22.12/24/26), happy-dom, Playwright browsers, React Server, Vercel Edge, workerd, Bun, and Deno—replacing tiny `runtimes/*` smoke tests and fixing jobs that weren’t actually exercising browsers or coverage. Tests are split by concern with `*.node.test.ts` / `*.browser.test.ts` suffixes instead of inline `skipIf`; **CONTRIBUTING.md** / **AGENTS.md** document no type assertions, `get-it/mock` (no module mocks), and integration rules. Browser upload tests move to a real local XHR server (`browserUpload.browser.test.ts`); packaging/CJS checks live in `test:packaging`. A new **Live API** workflow runs `*.integration.test.ts` against the real Sanity API on every PR (non-fork) and nightly across five runtimes.
> 
> **Client fixes (patch):** Live `/check/cors` uses `pickBaseFetch(config)` like EventSource so `resolveFetch`/`proxy` apply. `delete()` / `mutate()` default return types are `MultipleMutationResult` (runtime unchanged). `eventsource` is bumped to **≥5.1.0** for `lastEventId` on workerd; EventSource error handling is tightened. Media Library `assets.upload()` unwraps `{asset}` and exports `MediaLibraryAssetDocument`. `createVersion()` warning is reworded. README notes Cloudflare Workers `compatibility_date` for SSE.
> 
> **Tooling:** `knip`, `typecheck`, expanded scripts; CI drops build-artifact coupling for source-aliased runtime jobs; `test-happy-dom` / `test-react-server` added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b354c9f7942b4e77fe3c96e5931b4e179d8763b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bjoerge (dismissed):
Thank you so much for this @rexxars, this will make it so much easier to contribute here.
Also realized while reviewing that we're on pnpm now, yay.

🎉 ❤️

PR #1288, opened 2026-08-17
refactor: drop nanoid

rexxars:
### Description

`nanoid` was used a single location, to generate release IDs. While the dependency is quite small, it's still a dependency. Since we're only using it to generate a specific, 8-character long ID with a given alphabet, we don't need the configuration of nanoid - so this inlines a variant doing what we need. It takes about ~500ms to generate a _million_ iterations on my macbook pro, and `getRandomValues()` gets executed only once per ID generation in about 78% of cases, otherwise it takes two calls.

### What to review

Looks okay?

### Testing

Relying on existing tests here.

---

> [!NOTE]
> **Low Risk**
> Small dependency swap with the same ID shape and call sites; behavior change risk is limited to randomness implementation details covered by existing tests.
> 
> **Overview**
> Removes the **`nanoid`** dependency and inlines release ID generation in **`generateReleaseId`** (`createVersionId.ts`).
> 
> **`generateReleaseId`** still returns 8 characters from the same alphanumeric alphabet, but now builds them with **`crypto.getRandomValues()`** (rejecting bytes that map outside the alphabet via `byte & 63`) instead of **`nanoid`’s `customAlphabet`**. **`createRelease`** continues to use this helper for default **`releaseId`** values; lockfile entries for **`nanoid@6.0.1`** are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1922fb6d750d8cbb1b891d03bea0b5b5c65c6459. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,6 @@
   "dependencies": {
     "eventsource": "^5.1.0",
     "get-it": "^9.5.0",
-    "nanoid": "^6.0.1",
     "obug": "^2.1.4",
     "rxjs": "^7.8.2"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,9 +14,6 @@ importers:
       get-it:
         specifier: ^9.5.0
         version: 9.5.0(vitest@4.1.10)
-      nanoid:
-        specifier: ^6.0.1
-        version: 6.0.1
       obug:
         specifier: ^2.1.4
         version: 2.1.4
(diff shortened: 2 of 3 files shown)


stipsan (approved):
Nice, less deps = less problems

PR #1295, opened 2026-08-21
fix(functions): enable sync and async function invokes

macdonst:
### Description

Now supports async invoke:

```javascript
const result = await client.functions.invoke('invoke-too', {
  event: {data: {_id: 'movie_118340'}}
})
```

and sync invoke:

```javascript
const result = await client.functions.invoke('invoke-too', {
  event: {data: {_id: 'movie_118340'}}
}, {sync: true})
```

### What to review

Please make sure I did the types correctly.

### Testing

Unit tests updated and manually tested against prod.

---

> [!NOTE]
> **Medium Risk**
> Public API behavior change: existing callers of `functions.invoke()` no longer wait for a return value unless they pass `{sync: true}`. Not a security-critical path, but it is a silent default-behavior shift.
> 
> **Overview**
> **`client.functions.invoke()` is now async by default.** The call resolves with `undefined` as soon as the service accepts it. Pass `{sync: true}` to wait for a return value (still only `sanity.function.pubsub`).
> 
> Async invoke also works for `sanity.function.durable` and `sanity.function.queue`. Sync on those types is rejected before hitting the Functions service. The invoke URL gets `?sync=true` only when requested.
> 
> Overloads type async as `undefined`, sync as `R`, and a non-literal `sync` as `R | undefined`. An explicit type argument without `{sync: true}` keeps the old `R | undefined` shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b11cee0a9fa9995ce0b887deeaca80441029c934. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/functions-invoke-sync.md
+++ b/.changeset/functions-invoke-sync.md
@@ -1,0 +1,30 @@
+---
+'@sanity/client': minor
+---
+
+feat: invoke functions synchronously or asynchronously
+
+`client.functions.invoke()` now takes an `options` argument and accepts `sanity.function.durable`
+and `sanity.function.queue` functions alongside `sanity.function.pubsub`.
+
+By default the invocation is queued: the call resolves with `undefined` as soon as the Functions
+service accepts it, without waiting for the function to run. Pass `{sync: true}` to keep the
+request open until the function finishes and resolve with its return value — that adds a
+`?sync=true` query parameter to the invoke request, and remains limited to
+`sanity.function.pubsub`.
+
+```ts
+// Queued — resolves once accepted, with no return value.
+await client.functions.invoke('my-func', {event: {data: {hello: 'world'}}})
+
+// Inline — resolves with whatever the function returns.
+const result = await client.functions.invoke<Result>(
+  'my-func',
+  {event: {data: {hello: 'world'}}},
+  {sync: true},
+)
+```
+
+The return type follows the option: a call with `{sync: true}` resolves with `R`, while an async
+call resolves with `undefined`. Callers who pass an explicit type argument without `{sync: true}`
+keep the previous `R | undefined` shape.
--- a/.changeset/functions-invoke-sync.md
+++ b/.changeset/functions-invoke-sync.md
@@ -7,17 +7,17 @@ feat: invoke functions synchronously or asynchronously
 `client.functions.invoke()` now takes an `options` argument and accepts `sanity.function.durable`
 and `sanity.function.queue` functions alongside `sanity.function.pubsub`.
 
-By default the invocation is queued: the call resolves with `undefined` as soon as the Functions
+By default the invocation is queued: the call resolves with `undefined` as soon as the
 service accepts it, without waiting for the function to run. Pass `{sync: true}` to keep the
 request open until the function finishes and resolve with its return value — that adds a
 `?sync=true` query parameter to the invoke request, and remains limited to
 `sanity.function.pubsub`.
 
 ```ts
-// Queued — resolves once accepted, with no return value.
+// Async — resolves once accepted, with no return value.
 await client.functions.invoke('my-func', {event: {data: {hello: 'world'}}})
 
-// Inline — resolves with whatever the function returns.
+// Sync — resolves with whatever the function returns.
 const result = await client.functions.invoke<Result>(
   'my-func',
   {event: {data: {hello: 'world'}}},
--- a/src/functions/FunctionsClient.ts
+++ b/src/functions/FunctionsClient.ts
@@ -2,7 +2,7 @@ import {lastValueFrom, type Observable} from 'rxjs'
 
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
 import type {HttpRequest} from '../types'
-import {_invoke, type InvokeFunctionRequest} from './invoke'
+import {_invoke, type InvokeFunctionOptions, type InvokeFunctionRequest} from './invoke'
 
 /** @public */
 export class ObservableFunctionsClient {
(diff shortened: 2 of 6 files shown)


.changeset/functions-invoke-sync.md
```diff
@@ -0,0 +1,30 @@
...
+feat: invoke functions synchronously or asynchronously
+
+`client.functions.invoke()` now takes an `options` argument and accepts `sanity.function.durable`
+and `sanity.function.queue` functions alongside `sanity.function.pubsub`.
+
+By default the invocation is queued: the call resolves with `undefined` as soon as the Functions
+service accepts it, without waiting for the function to run. Pass `{sync: true}` to keep the
+request open until the function finishes and resolve with its return value — that adds a
+`?sync=true` query parameter to the invoke request, and remains limited to
+`sanity.function.pubsub`.
```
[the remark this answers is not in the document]
macdonst: I'm not worried about this as no customers have deployed a `pubsub` function. This is why I want to get this out ASAP.

.changeset/functions-invoke-sync.md
```diff
@@ -0,0 +1,30 @@
+---
+'@sanity/client': minor
+---
+
+feat: invoke functions synchronously or asynchronously
+
+`client.functions.invoke()` now takes an `options` argument and accepts `sanity.function.durable`
+and `sanity.function.queue` functions alongside `sanity.function.pubsub`.
+
+By default the invocation is queued: the call resolves with `undefined` as soon as the Functions
```
filmaj: Should we be mentioning internal Sanity things like 'Functions service'? I also dislike using the term 'queued' with respect to the invoke request - given there's a funciton type called a queue imminent, I think it muddies the picture. "By default the invocation is asynchronous and returns immediately" is clearer IMO

.changeset/functions-invoke-sync.md
```diff
@@ -0,0 +1,30 @@
...
+service accepts it, without waiting for the function to run. Pass `{sync: true}` to keep the
+request open until the function finishes and resolve with its return value — that adds a
+`?sync=true` query parameter to the invoke request, and remains limited to
+`sanity.function.pubsub`.
+
+```ts
+// Queued — resolves once accepted, with no return value.
```
filmaj: The comment here uses 'queued' again. The next comment below uses 'inline'. I don't like either of these terms to refer to the differing sync/async behaviour. Why not stick to sync and async?

src/functions/FunctionsClient.ts
```diff
@@ -17,16 +17,35 @@ export class ObservableFunctionsClient {
    * Invoke a deployed function by its blueprint name.
    *
    * The name is resolved within the stack given by `stackId` on the request or
-   * the client config. Passes the function's return value once it finishes.
+   * the client config. Queues the invocation and emits `undefined` as soon as
```
filmaj: "queues" 👀

src/functions/invoke.ts
```diff
@@ -37,6 +45,18 @@ export interface InvokeFunctionRequest {
   signal?: AbortSignal
 }
 
+/** @public */
+export interface InvokeFunctionOptions {
+  /**
+   * Wait for the function to finish and resolve with its return value.
+   *
+   * Defaults to `false`: the invocation is queued, the request resolves as soon
```
filmaj: 'queued' 👀

src/functions/FunctionsClient.ts
```diff
@@ -17,16 +17,35 @@ export class ObservableFunctionsClient {
...
+    functionName: string,
+    request?: InvokeFunctionRequest,
+    options?: InvokeFunctionOptions & {sync?: false},
+  ): Observable<undefined>
+  invoke<R = unknown>(
+    functionName: string,
+    request: InvokeFunctionRequest | undefined,
+    options: InvokeFunctionOptions & {sync: true},
+  ): Observable<R>
+  invoke<R = unknown>(
```
filmaj, line 38: Isn't this signature the same as the next one below?

src/functions/FunctionsClient.ts
```diff
@@ -44,23 +63,45 @@ export class FunctionsClient {
...
+    functionName: string,
+    request?: InvokeFunctionRequest,
+    options?: InvokeFunctionOptions & {sync?: false},
+  ): Promise<undefined>
+  invoke<R = unknown>(
+    functionName: string,
+    request: InvokeFunctionRequest | undefined,
+    options: InvokeFunctionOptions & {sync: true},
+  ): Promise<R>
+  invoke<R = unknown>(
```
filmaj, line 94: Same here, I don't think this signature is needed as it is a duplicate of the one below it.

filmaj (changes requested):
I think some TS function signatures are duplicates. I also don't like the comments that refer to sync invokes as being 'queued', and async invokes as being 'inline'. I think that's confusing terminology.

src/functions/FunctionsClient.ts
macdonst: Preface: I'm not a TS expert.

I think we need both signatures. The first is an overload signature which is public and participates in overload resolution. The second duplicate signature is an implementation signature. Once a function has any overloads, TypeScript hides the implementation signature from callers. If you build the client and look at the signatures in `dist/index.node.d.ts` you'll see that `FunctionsClient` and `ObservableFunctionsClient` only have 3 method signatures.

src/functions/FunctionsClient.ts
macdonst: Same reply as above.

src/functions/FunctionsClient.ts
filmaj: Oh nice. TIL! Sorry about the false alarm

filmaj (approved):

PR #1304, opened 2026-08-27
fix!: throw when querying drafts perspective with the API-CDN

jordanl17:
> [!NOTE]
> Hold this for the next major release of `@sanity/client`. It is a valid SemVer break and should land with that major, not as a dedicated v9 for this change alone.

Fixes [SAPP-2685](https://linear.app/sanity/issue/SAPP-2685/js-client-returns-a-warning-for-a-gradient-error).

## Reproduction

The JS client still only warned when `perspective: 'drafts'` (or `'previewDrafts'`) was combined with `useCdn: true`. Gradient rejects that combination on the API-CDN; the client used to swallow it, switch to the Live API, and print:

> The Sanity client is configured with the `perspective` set to `drafts` or `previewDrafts`, which doesn't support the API-CDN. The Live API will be used instead.

That matches the original report. We own this path in `@sanity/client`.

## Before / after

Covered in `test/client/cdnPerspective.test.ts`.

**Before:** `createClient({useCdn: true, perspective: 'drafts'}).fetch(...)` warned and queried `api.sanity.io`.
**After:** the same call throws:

`The Sanity client is configured with the perspective set to drafts or previewDrafts, which does not support the API-CDN. Set useCdn: false.`

Still allowed:
- `useCdn: false` + `drafts` (Live API)
- `useCdn: true` + `published` / `raw` (CDN)
- per-fetch `{useCdn: false}` on a drafts client
- stacked perspectives that do **not** include `drafts` still fall back to the Live API with a warning (same as today, so release stacks keep working)

This remains a major (`fix!`) because existing queries that succeeded with a warning will now throw. The old warning asked callers to hide it; it was not a documented deprecation with a removal window.

<div><a href="https://cursor.com/agents/bc-783c0b5d-79b0-4dbd-b821-568a26115863?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-783c0b5d-79b0-4dbd-b821-568a26115863&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/pr-1304.md
+++ b/.changeset/pr-1304.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/client': major
+---
+
+feat!: throw when querying drafts perspective with the API-CDN
--- a/.changeset/pr-1304.md
+++ b/.changeset/pr-1304.md
@@ -3,4 +3,4 @@
 '@sanity/client': major
 ---
 
-fix!: throw when querying drafts perspective with the API-CDN
+feat!: throw when querying drafts perspective with the API-CDN
--- a/.changeset/pr-1304.md
+++ b/.changeset/pr-1304.md
@@ -3,4 +3,4 @@
 '@sanity/client': major
 ---
 
-feat!: throw when querying drafts perspective with the API-CDN
+fix!: throw when querying drafts perspective with the API-CDN
--- a/src/config.ts
+++ b/src/config.ts
@@ -41,6 +41,27 @@ export function validateApiPerspective(
   }
 }
 
+/**
+ * Gradient rejects `drafts` and `previewDrafts` on the API-CDN. Stacked
+ * perspectives that include those values are treated the same. Other stacked
+ * perspectives still fall back to the Live API with a warning.
+ *
+ * @internal
+ */
+export function perspectiveConflictsWithCdn(perspective: ClientPerspective): boolean {
+  if (perspective === 'previewDrafts' || perspective === 'drafts') {
+    return true
+  }
+  if (Array.isArray(perspective)) {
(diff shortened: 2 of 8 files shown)


rexxars:
@jordanl17 Is there a sense of urgency behind this? We just recently released a new major, and getting all surfaces and peer dependency ranges aligned has already taken quite a bit of time. Obviously that's a situation we'd like to improve/make easier, but just questioning whether or not this is really the right time to do another major. I personally do not have the time to cascade this throughout repos. If there's no immediate urgency, I'd like to schedule this under the v9 milestone

rexxars (changes requested):
Rejecting this based on my previous comment to avoid it getting merged without taking it into account - happy to re-review if urgent.

.changeset/pr-1304.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/client': major
```
stipsan, line 3: lol clanker, we are not doing a major for this

stipsan (changes requested):

PR #1323, opened 2026-09-03
feat: configure transparent OAuth token refresh via auth.oauth

joshuaellis:
## Description

Apps authenticating with short-lived OAuth access tokens would have to catch 401s, refresh, and re-create clients themselves which is annoying for everyone involved.

This lets the client own that instead: configure `auth.oauth` with an `OAuthTokenSetup` and requests, live streams, and uploads stay authenticated across token rotations without the app noticing. When a token genuinely can't be refreshed because the session is over then the app gets a single `onAuthError` signal to send the user back through login, instead of scattered 401s.

---

> [!NOTE]
> **High Risk**
> Touches authentication across HTTP, uploads, and long-lived SSE connections, including refresh deduplication and 401/reconnect semantics that affect session security.
> 
> **Overview**
> Adds **`auth.oauth`** (`OAuthTokenSetup` with `getToken`, `refresh`, optional `getExpiresAt` / `onAuthError`) as an alternative to a static `token`, with config validation that the two cannot be combined.
> 
> A new **`oauthRefreshHandler`** wraps the normal request pipeline: proactive refresh before expiry, **single-flight** `refresh()` across concurrent callers, **one automatic retry** on 401 for fetch-based APIs, and **`onAuthError`** when refresh fails. The same setup is wired into paths that skip the default handler—**asset uploads** (fetch and XHR), **`listen` / `live` EventSource** (Bearer resolved per fetch, including reconnects), plus **401-driven reconnect** with token refresh and **`Last-Event-ID`** resume so SSE can continue after auth recovery.
> 
> Uploads **do not auto-retry** after 401 (body may be consumed); they still refresh so a caller retry gets a new token.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 371fd57f9cbaf16159c5021c1a6d0403b01075bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -2,6 +2,7 @@ import {EventSource} from 'eventsource'
 import {Observable, throwError} from 'rxjs'
 import {filter, map} from 'rxjs/operators'
 
+import {getOAuthRefresher, getOAuthTokenSetup} from '../http/oauthRefreshHandler'
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
 import {
   type Any,
@@ -166,23 +167,25 @@ export function _connectListenEventSource<TEvent extends {type: string}>(
   const {token, withCredentials, headers: configHeaders} = config
 
   const headers: Record<string, string> = {}
-  if (token) {
+  if (typeof token === 'string') {
     headers.Authorization = `Bearer ${token}`
   }
   if (configHeaders) {
     Object.assign(headers, configHeaders)
   }
+  const tokenSetup = getOAuthTokenSetup(token)
 
   const initEventSource = () =>
     new EventSource(uri, {
       fetch: resolveEventSourceFetch(config, {
         headers: Object.keys(headers).length ? headers : undefined,
+        tokenSetup,
         withCredentials,
       }),
     })
 
   return connectEventSource(initEventSource, listenFor).pipe(
-    reconnectOnConnectionFailure(),
+    reconnectOnConnectionFailure(tokenSetup && getOAuthRefresher(tokenSetup)),
     filter((event) => listenFor.includes(event.type)),
     map((event) => ({
       type: event.type,
--- a/src/data/listen.ts
+++ b/src/data/listen.ts
@@ -1,6 +1,6 @@
 import {EventSource} from 'eventsource'
 import {Observable, throwError} from 'rxjs'
-import {filter, map} from 'rxjs/operators'
+import {filter, map, tap} from 'rxjs/operators'
 
 import {getOAuthRefresher, getOAuthTokenSetup} from '../http/oauthRefreshHandler'
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
@@ -174,17 +174,24 @@ export function _connectListenEventSource<TEvent extends {type: string}>(
     Object.assign(headers, configHeaders)
   }
   const tokenSetup = getOAuthTokenSetup(token)
+  let lastEventId: string | undefined
 
   const initEventSource = () =>
     new EventSource(uri, {
       fetch: resolveEventSourceFetch(config, {
         headers: Object.keys(headers).length ? headers : undefined,
+        lastEventId,
         tokenSetup,
         withCredentials,
       }),
     })
 
   return connectEventSource(initEventSource, listenFor).pipe(
+    tap((event) => {
+      if ('id' in event && typeof event.id === 'string' && event.id) {
+        lastEventId = event.id
+      }
+    }),
     reconnectOnConnectionFailure(tokenSetup && getOAuthRefresher(tokenSetup)),
     filter((event) => listenFor.includes(event.type)),
     map((event) => ({
--- a/src/data/live.ts
+++ b/src/data/live.ts
@@ -4,6 +4,7 @@ import {catchError, mergeMap, Observable, of, throwError} from 'rxjs'
 import {finalize, map} from 'rxjs/operators'
 
 import {CorsOriginError} from '../http/errors'
+import {getOAuthRefresher, getOAuthTokenSetup} from '../http/oauthRefreshHandler'
 import type {ObservableSanityClient, SanityClient} from '../SanityClient'
 import type {
   InitializedClientConfig,
@@ -13,6 +14,7 @@ import type {
   LiveEventReconnect,
   LiveEventRestart,
   LiveEventWelcome,
+  OAuthTokenSetup,
   SyncTag,
 } from '../types'
 import {isRecord} from '../util/isRecord'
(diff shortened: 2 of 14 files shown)


src/data/listen.ts
```diff
@@ -166,23 +167,25 @@ export function _connectListenEventSource<TEvent extends {type: string}>(
...
       fetch: resolveEventSourceFetch(config, {
         headers: Object.keys(headers).length ? headers : undefined,
+        tokenSetup,
         withCredentials,
       }),
     })
 
   return connectEventSource(initEventSource, listenFor).pipe(
-    reconnectOnConnectionFailure(),
+    reconnectOnConnectionFailure(tokenSetup && getOAuthRefresher(tokenSetup)),
```
bjoerge, line 188: I think this will break resumability as it will trigger a new listener request, which instantiates a new EventSource rather than reusing the existing one, bypassing the existing EventSource's built-in error/retry mechanism (which sends the Last-Event-ID header on reconnect, telling the server where to resume from).

src/data/reconnectOnConnectionFailure.ts
```diff
@@ -37,6 +55,25 @@ export function reconnectOnConnectionFailure<T>(): OperatorFunction<T, T | {type
...
+            from(refreshAuth()).pipe(
+              // Surface the original connection error, not the refresh failure
+              // — `onAuthError` has already fired inside the refresher. Placed
+              // before `mergeMap` so it only catches the refresh promise, never
+              // errors from the resubscribed stream.
+              catchError(() => throwError(() => err)),
+              mergeMap(() => caught),
+            ),
+          )
+        }
```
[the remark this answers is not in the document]
joshuaellis: Focussing on the SSE 401 path not re-reading `getToken()` before refreshing, so it isn't consistent with the request handler and upload paths.

I don't think it can revoke a session in practice. Concurrent 401s share one refresh via the single-flight refresher. A later stale 401 calls `refresh()` again, and if `refresh()` reads the current refresh token from storage it just rotates a valid token. That wastes a rotation but is harmless. You'd only trip reuse detection if `refresh()` replays a token it captured earlier, which the client has no control over.

joshuaellis:
From slack: lets explore proactively refreshing the token before it expires (instead of waiting for a 401).

src/data/listen.ts
joshuaellis: I've had a look at producing tests which hopefully disproves your concern - let me know if i've missed a case worth having!

test/resumability.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  expect(await firstValueFrom(client.listen('*').pipe(take(2), toArray()))).toHaveLength(2)
+
+  const requests = getActiveMock().getRequests()
+  expect(requests.map((r) => r.headers.get('authorization'))).toEqual([
+    'Bearer expired',
+    'Bearer expired',
+    'Bearer fresh',
+  ])
+  // Last-Event-ID survives the lib's own reconnect (2nd request) but not ours (3rd)
+  expect(requests.map((r) => r.headers.get('last-event-id'))).toEqual([null, 'evt-1', null])
```
ryanbonial: @claude Will this continue to work OK if it resumes and then the last event ID is null? Does it matter if we miss events?

joshuaellis: I think you have to ask @copilot  here lol

ryanbonial (dismissed):

src/types.ts
```diff
@@ -132,7 +157,7 @@ export interface ClientConfig {
   dataset?: string
   /** @defaultValue true */
   useCdn?: boolean
-  token?: string
+  token?: string | OAuthTokenSetup
```
bjoerge: this is potentially breaking, no? a consumer doing e.g. `Bearer ${client.config().token}`. Should we consider a different field for OauthTokenSetup?

src/data/reconnectOnConnectionFailure.ts
```diff
@@ -37,6 +55,25 @@ export function reconnectOnConnectionFailure<T>(): OperatorFunction<T, T | {type
...
+        if (
+          refreshAuth &&
+          err instanceof ConnectionFailedError &&
+          err.status === 401 &&
+          Date.now() - lastAuthRetryAt >= AUTH_RETRY_WINDOW
+        ) {
+          lastAuthRetryAt = Date.now()
+          return concat(
+            of({type: 'reconnect' as const}),
+            from(refreshAuth()).pipe(
```
bjoerge: not a huge deal I think, but this calls `refreshAuth()` eagerly. Would be better to use  `defer(refreshAuth)` here to make it call refreshAuth at subscription time.

src/data/resolveEventSourceFetch.ts
```diff
@@ -50,19 +68,29 @@ export function resolveEventSourceFetch(
...
     // declare (`mode`, `cache`) survive the spread and reach whichever
     // fetch implementation is effective.
     const mergedInit: FetchInit = {...init}
-    if (extraHeaders) {
+    if (extraHeaders || tokenSetup || lastEventId) {
       const headers = new Headers(init?.headers)
-      for (const [key, value] of Object.entries(extraHeaders)) {
-        headers.set(key, value)
+      if (lastEventId && !headers.has('last-event-id')) {
+        headers.set('Last-Event-ID', lastEventId)
```
bjoerge, line 85: EventSource instances also tracks Last-Event-ID internally, how will these interact? We need to be extra careful here so we don't lose events or get duplicated delivery. Find it a bit hard to judge whether that's a real concern though 😅

src/data/resolveEventSourceFetch.ts
```diff
@@ -71,7 +99,14 @@ export function resolveEventSourceFetch(
...
     // get-it's `FetchResponse` is a structural superset of the package's
     // `FetchLikeResponse`, so it can be handed over as-is.
-    return baseFetch(typeof url === 'string' ? url : url.href, mergedInit)
+    const response = baseFetch(typeof url === 'string' ? url : url.href, mergedInit)
+    // Returning a promise from an async function attaches its rejection
+    // handler one microtask later (thenable adoption), and workerd's
+    // unhandled-rejection tracker flags a rejected promise in that gap.
+    // Attach a no-op handler synchronously; the rejection still propagates
+    // through the async return to the `eventsource` package's catch.
+    response.catch(() => {})
```
bjoerge: This feels like a hack to me. Any way we can avoid this?

src/types.ts
joshuaellis: Yeah it would break TS because it _could_ be an object instead of a string. I've updated to a separate field – lmk what you think.

src/data/reconnectOnConnectionFailure.ts
joshuaellis: TIL! Thanks ✅

src/data/resolveEventSourceFetch.ts
joshuaellis: Agree, reworked.

src/data/resolveEventSourceFetch.ts
joshuaellis: > The wrapper only sets the header when the package hasn't (!headers.has('last-event-id')), and both values come from the same source (#lastEventId, surfaced as message.lastEventId). So on a live instance's own reconnects the package's header wins; the client-tracked value only seeds a fresh instance. Added a test in 0be6f0e for the case that would go wrong: fresh instance seeded with evt-1, receives evt-2, drops, and its own reconnect sends evt-2.

I've (claude) tried to add tests to cover this. Beyond that i'm not sure how else we can test this – open to ideas.

src/data/live.ts
```diff
@@ -96,6 +98,10 @@ export class LiveClient {
     if (configHeaders) {
       Object.assign(eventSourceHeaders, configHeaders)
     }
+    // An OAuth token setup can't be baked into the headers — it is resolved
```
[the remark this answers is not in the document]
joshuaellis: @copilot is this true?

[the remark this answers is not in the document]
joshuaellis: it was true.

PR #1334, opened 2026-09-15
feat: add per-resource schema, query and projection registries

ryanbonial:
### Description

Adds three type registries alongside the existing `SanityQueries`, each keyed by resource:

```ts
declare global {
  interface SanitySchemasByResource {}
  interface SanityQueriesByResource {}
  interface SanityProjectionsByResource {}
}
```

Each follows the pattern `SanityQueries` already uses: a global interface, plus an interface
exported from `@sanity/client` that inherits it. A generated file can register either way, and
neither form depends on module resolution.

Nothing in the client reads them. `client.fetch`, `ClientReturn` and `ClientReturnStega` are
unchanged and keep resolving through the flat registry.

### Why

The flat `SanityQueries` registry cannot express two datasets whose schemas give the same query
text different result types. An app reading from a `production` and a `staging` dataset where
`post.title` is a string in one and a number in the other gets one type for both. Keying by
resource is what fixes that, and it is the agreed direction for multi-resource TypeGen across
the App SDK and other consumers.

A resource key is the same string the App SDK already builds for its runtime store cache:
`projectId.dataset` for a dataset, no prefix. Generated types using that same format means
nobody debugging has to hold two key shapes in their head.

Projections carry a document-type level that queries do not. A projection applies to whichever
document the caller's handle names, so the same projection text resolves differently per
document type as well as per resource.

Schemas have only ever been registered in the `groq` module, so this moves them here, where
the query registry already lives and where consumers who call `client.fetch` without a `groq`
dependency can reach them.

### Testing

`test/codegen/resourceRegistries.test-d.ts` adds 9 type tests: two resources whose schemas
disagree about `title`, registered through both the global and the module-augmentation form,
asserting that the same query text resolves to a different type for each.

`typecheck`, `lint`, `format:check` and `build` are clean, and `test/codegen/client.test-d.ts`
still passes its 14 tests. The build's 60 `ae-extra-release-tag` warnings are pre-existing, on
declarations carrying both `@public` and `@beta`.

---

> [!NOTE]
> **Low Risk**
> Additive TypeScript-only declarations with no runtime or fetch behavior changes; existing `ClientReturn` resolution is unchanged.
> 
> **Overview**
> Adds **three optional type registries** alongside the existing flat `SanityQueries`: `SanitySchemasByResource`, `SanityQueriesByResource`, and `SanityProjectionsByResource`. Each uses the same pattern as `SanityQueries`—empty global interfaces plus `@sanity/client` exports that inherit `globalThis`—so `sanity typegen` can register via globals or module augmentation without depending on module resolution.
> 
> Resource keys use **`projectId.dataset`**, matching the App SDK runtime cache. That lets the same GROQ query string map to **different result types per dataset** when schemas disagree; projections add a **document-type** level because projection text is resolved in the context of a specific handle.
> 
> The registries are **empty by default** and **not wired into** `client.fetch`, `ClientReturn`, or `ClientReturnStega`; those still use the flat query registry. This is infrastructure for consumers (e.g. App SDK hooks) that do resource-aware lookups.
> 
> Adds **type-only tests** (`test/codegen/resourceRegistries.test-d.ts` and `test/packaging/resourceRegistries.test-d.ts`) covering per-resource query typing, schema unions, projections, and that built `.d.ts` still exposes the bridged interfaces.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47ce4647585bd85aacc0ab58b8c71128f473b64b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cngonzalez (commented):
Generally looks good to me but I have a question about the schema types specifically -- I don't believe current Typegen nests schemas under any namespace (IIRC); does nesting schemas under the client here affect scanning or resolving types for things like `defineQuery`?

ryanbonial:
@cngonzalez Yes, the current Typegen exports schema types directly. Those will stay as the same and still be available if needed. The new type gives us a way to look up the generated schema union for a given project/dataset:
```ts
// Existing exported types remain above.
declare global {
  interface SanitySchemasByResource {
    'project123.production': AllSanitySchemaTypes
  }
}
```
These won't be used for `defineQuery` scanning. Typegen extracts the query string and evaluates it against the extracted schema without these registries.

cngonzalez (approved):
Then I'm all right with it, let's go 🚀

